PR #2, opened 2020-03-18
Added --size and: fixed typO ?

izznogooood:
Added a size option.

I'm not sure but maybe "snap connect qemu-virgil:removable-media" was supposed to be "snap connect qemu-virgil:kvm"

/ Anders


flexiondotorg:
I made some changes which caused this PR to conflict. I've added details about connecting KVM to the README.md, thank for that :+1: 

As for disk size, I plan to add that via the VM configuration file :smiley:

PR #4, opened 2020-03-19
Enable SSH to the guest machine

stuh84:
This PR enables SSH to be port forwarded to the guest machine, with a default host port of TCP2222. This will then forward to the guest on port 22.

This also adds the ability to override the port that is forwarded (if it clashes with another port used locally).

Let me know if anything needs to change on it, either code style, documentation, or any problems found.


flexiondotorg:
Thanks for the idea and for contributing. After our chat I added this feature via https://github.com/wimpysworld/quickemu/commit/0a1a7b3851b65aeb880113781a79fbc83835789d

Issue #22, opened 2020-04-05
Allow for cpu core usage in os configuration


SneWs:
Similar to having RAM configurable, I would find it useful to be able to define the amount of cores to use for a OS configuration. For example:

`cpu_cores=8`

In the os configuration file.

flexiondotorg:
Closed via https://github.com/wimpysworld/quickemu/commit/8618421abb0a1cdadbe5e10aed139eef96c91011

PR #24, opened 2020-04-05
Fix #21 configurable RAM and #22 configurable cores

### LINKED ISSUE #21: Allow for RAM configuration in os conf file
When defining and configuring a OS to be run, one can today define disk size etc. But RAM usage is not configurable. Would be nice to be able to define something like:

`ram=8G`

in the config file.

SneWs:
This enables one to override the automatic selection of RAM and CPU resources.

### DIFF
--- a/quickemu
+++ b/quickemu
@@ -279,23 +279,31 @@ function vm_boot() {
   fi
 
   local CORES_VM="1"
-  local CORES_HOST=$(nproc --all)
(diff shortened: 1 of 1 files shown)


SneWs:
That should do it, I think.

flexiondotorg:
Thank you for that! :smiley:

PR #28, opened 2020-04-13
Allow using installed version of qemu

MatejSpindler:
Some distributions like ArchLinux provide properly configured QEMU, with working virgil. On such installs it's pointless to install additional version of QEMU.

This pull request adds option "--system-qemu" to run VM's with distribution provided version of QEMU.


flexiondotorg:
Thanks for contributing!

PR #29, opened 2020-04-13
find free port for spice

marosg42:
allows to run more VMs in parallel


flexiondotorg:
Thanks for contributing!

PR #40, opened 2020-06-24
Enable arbitrary port forwarding via config file

ymauray:
In order to run Home Assistant in a VM, I need certain ports to be forwarded from my machine to my VM, so I can access the web interface or the embeded MQTT server.
I've looked at how th USB passthrough was made and tried to replicate that for forwards.


flexiondotorg:
Nice feature! Thanks for contributing :smiley:

PR #104, opened 2021-10-14
Adding impish to the Ubuntu releases for quickget

philclifford:
This seems to be all there is to do so did this noddy one-liner as a tiny token of appreciation for the incredible work getting the release done.


philclifford:
Yannik's updated the Readme too in #106 so closing this to save time in the morning

PR #121, opened 2021-10-18
Adding support for openSUSE MicroOS and openSUSE Kubic

apinter:
Adding openSUSE [MicroOS](https://microos.opensuse.org/) and Kubic(https://kubic.opensuse.org/) support to `quickget`.


apinter:
I see. Will remove Kubic and leave in only MicroOS as it does have an immutable desktop role with GNOME and Plasma. brb

PR #174, opened 2021-10-29
Android Support

Guy-Sowden:
Added support for x86 versions of android.
https://www.android-x86.org/


matteoturini:
Add 9.0-r2 support

PR #185, opened 2021-10-31
Added support for Zorin OS

philclifford:
The Zorin iso does optional self-validation on boot it seems, but no sign of a hash to check.
quickemu boots ok but (at least for me ) you need to pick the "safe graphics" options when the installer iso boots or you get a non-working black screen.


philclifford:
> ... (at least for me ) you need to pick the "safe graphics" options when the installer iso boots or you get a non-working black screen.
Now we are linux this seems to be sorted .  NB the "education isos want ~22G in the disk so that'll need handling manually on in a post-lunch update

flexiondotorg:
Thanks very much!

PR #190, opened 2021-11-01
Added support for Regolith Linux

philclifford:
focal and hirsute versions of Release 1.6.0
plus hirsute and impish based pre-release 2.0.0


flexiondotorg:
Thanks! 🙂

PR #198, opened 2021-11-02
fix for shortcut presumption

### LINKED ISSUE #189: quickemu --shortcut fails in less than clear way when ~/.local/applications doesn't exist
#### Expected behaviour

If I haven't yet created *~/local/applications*, `quickemu --shortcut` should tell me to do so, or do it for me.

#### Actual behaviour

Quickemu tells me that the shortcut it was supposed to create doesn't exist, then that it created it‽ I find no shortcut in my start menu, and realize that the folder to contain it doesn't exist.

#### Steps to reproduce the behaviour

Outline the step to reproduce the issue you've encountered.

#### Quickemu output

```bash
➜  quickemu --vm macos-monterey.conf --shortcut
/usr/bin/quickemu: line 937: /home/dale/.local/share/applications/macos-monterey.desktop: No such file or directory
Created macos-monterey.desktop file

➜  ls ~/.local/share/applications
ls: cannot access '/home/dale/.local/share/applications': No such file or directory

✗  mkdir ~/.local/share/applications

➜  quickemu --vm macos-monterey.conf --shortcut
Created macos-monterey.desktop file
```

#### Linux Distribution & Kernel

Run the following and paste the output.

```bash
➜  lsb_release --all
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 21.10
Release:	21.10
Codename:	impish

~ took 8s 
➜  uname -a
Linux octavia 5.13.0-20-generic #20-Ubuntu SMP Fri Oct 15 14:21:35 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

philclifford:
fixes #189

### DIFF
--- a/quickemu
+++ b/quickemu
@@ -933,7 +933,12 @@ function vm_boot() {
 }
 
 function shortcut_create {
+  local dirname="${HOME}/.local/share/applications"
   local filename="${HOME}/.local/share/applications/${VMNAME}.desktop"
+
+  if [ ! -d ${dirname} ]; then
+    mkdir -p ${dirname}
+  fi
   cat << EOF > "${filename}"
 [Desktop Entry]
 Version=1.0


quickemu
```diff
@@ -933,7 +933,12 @@ function vm_boot() {
 }
 
 function shortcut_create {
+  local dirname="${HOME}/.local/share/applications"
   local filename="${HOME}/.local/share/applications/${VMNAME}.desktop"
```
omartijn, line 937: To keep in line with DRY, I'd suggest using the new `dirname` variable here as the base, and then adding `/${VMNAME}.desktop` to it to get the final `filename`.

omartijn (changes requested):

quickemu
philclifford: Good suggestion.  I'm in two minds though: for me this is a little too far in the direction of premature refactoring since

- it doesn't really pass the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))
- it makes the code slightly less obvious without any significant performance or security benefit
- it impacts code beyond the minimal scope of this patch.

I know I am overstating my counter-argument here (and would be happy to have a chat over a beer in [Discord](https://discord.gg/sNmz3uw) ), but my preference would be to consider this refactoring along with further possible DRY-cleaning in a separate issue/PR.

omartijn: I wouldn't call it _refactoring_ in this case. The `dirname` variable is only introduced in this PR and thus this is the first place where we _can_ use it to then compose it into the `filename`. 

Anyways, they're just my two cents. Feel free to ignore me if you disagree.

dwvisser:
I had just made my own version of this. 😄 This looks good though!

flexiondotorg:
Thanks!

PR #207, opened 2021-11-05
Add KDE Neon to quickget

daPhipz:
[KDE Neon](https://neon.kde.org) follows a rolling release model and provides several editions, each with a different level of testing vs stability (Unstable, Testing and User Edition).
On top, there is also a Developer Edition (Unstable plus some dev tools and libs preinstalled).

The reason I file this PR as a draft, is that I need help getting `zsync` to work. All files are located under https://files.kde.org/neon/images/user/current/ (replace "user" with the desired edition).
According to the `zsync` README in this directory, `files.kde.org` will always force a `https` connection, making it unusable for downloads with `zsync`. Instead, they suggest using a mirror - there is a nice webpage that shows you the nearest one at https://files.kde.org/neon/images/user/current/neon-user-current.iso.zsync.mirrorlist.

That's where I need help: Is there a way to parse this mirrorlist webpage in order to get the best mirror for the user's location? I could just hardcode to use one single mirror, but that would defeat the purpose of using mirrors in the first place.

I would be grateful for comments or PRs to my branch (`daPhipz/add-kde-neon`). Thanks!

### DIFF
--- a/quickget
+++ b/quickget
@@ -38,6 +38,7 @@ function pretty_name() {
     archlinux)          PRETTY_NAME="Arch Linux";;
     elementary)         PRETTY_NAME="elementary OS";;
     freebsd)            PRETTY_NAME="FreeBSD";;
+    kdeneon)           PRETTY_NAME="KDE Neon";;
     linuxmint-cinnamon) PRETTY_NAME="Linux Mint Cinnamon";;
     linuxmint-mate)     PRETTY_NAME="Linux Mint MATE";;
     linuxmint-xfce)     PRETTY_NAME="Linux Mint XFCE";;
@@ -132,6 +133,7 @@ function os_support() {
     freebsd \
     fedora \
     kali \
+    kdeneon \
     kubuntu \
     linuxmint-cinnamon \
     linuxmint-mate \
@@ -187,6 +189,13 @@ function releases_fedora(){
 function releases_kali() {
     echo latest \
     weekly
+}
+
+function releases_kdeneon() {
+    echo user \
+    testing \
+    unstable \
+    developer
 }
 
 function releases_linuxmint(){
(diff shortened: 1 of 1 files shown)


philclifford:
this works for me:

```
 wget -q -O- https://files.kde.org/neon/images/user/current/neon-user-current.iso.zsync.mirrorlist| grep 'neon-user-current.iso.zsync' neon-user-current.iso.zsync.mirrorlist | grep '>1.<'|cut -d\" -f 6
```

philclifford:
actually only need the second `grep` ( `grep '>1.<'` ) 
Of course if / when the mirrorlist output changes this may break. Might be worth adding a guard to abort if $URL doesn't conform to  a URL pattern.

daPhipz:
I think it does make sense to keep the first `grep` as a safety measure for exactly this reason.
There might be some other place on that website that matches `'>1.<'` in the future ;)

daPhipz:
Thank you very much @philclifford! After your suggestion, I started working on it myself. I incorporated your code and made you a co-author of this commit. Thanks again! :rocket:

flexiondotorg:
Thank you!

daPhipz:
@flexiondotorg While watching the first half of your stream as VOD (I tuned in later), I noticed that you were wondering why I only used `zsync_get` here and not `web_get`.
Is there a reason why you should use another method when the distros provide a zsync file (especially for rolling releases)?

philclifford:
>  Is there a reason why you should use another method when the distros provide a zsync file (especially for rolling releases)?

Just to summarise for non-Twitchers/Discordians : I think there are a few things to note currently regarding zsync

- you're right, for rolling/weekly/nightly ISOs it is absolutely the optimal choice to use zsync if available
- if you do use zsync you also need to add to the list output to be consumed by users and GUIs
- it turns out some distros don't ship zsync (yet?) in their repos

See you on a branch somewhere working on some tweaks to improve the Fedora 35 users' experience, and maybe a little refactoring so you only need to specify zsync in one place ....

PR #214, opened 2021-11-08
Fix 212 Fallback to wget if zsync not available

### LINKED ISSUE #212: Fall back to wget if zsync not available
@flexiondotorg mentioned on the stream a comment from @cowboygeneral that not every distro (Fedora 35) has zsync in the repo, and that we need to take care of that both in the gui and in quickget.

Suggestion: if zsync not found in path then set a guard that prevents zsync-only options from showing, listing or getting

Hopefully masking them out of the output of list then quickgui will just work.

philclifford:
fixes #212 
check if zsync is in the path and fallback to wget within zsync_get and list_csv
Thanks to @daPhipz for helpful discussion and contributions.

### DIFF
--- a/quickget
+++ b/quickget
@@ -632,8 +632,8 @@ function zsync_get() {
           rm "${DIR}/${OUT}.zs-old"
       fi
     else
-      echo "zsync not found so falling back to wget"
-      web_get ${ISO}
+      echo "INFO: zsync not found, falling back to wget"
+      web_get "${ISO}"
     fi
 }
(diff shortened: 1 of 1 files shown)


daPhipz:
I've filed a PR against your branch - great work, I really like it!

philclifford (commented):
Cherry-picked from @daPhipz contributions.

flexiondotorg:
Thanks!

PR #217, opened 2021-11-09
Fix broken network after bridge support in PR #160

navycatt:
This simply fixes the broken network after adding bridge support in PR https://github.com/wimpysworld/quickemu/pull/160

For qemu, it is apparently important that the network is defined earlier, and not at the end.

Issue https://github.com/wimpysworld/quickemu/issues/215.

### DIFF
--- a/quickemu
+++ b/quickemu
@@ -791,6 +791,15 @@ function vm_boot() {
     sleep 0.25
   fi
 
+  # Network mode
+  if [ -n "${bridge}" ]; then
+    # Enable bridge mode networking
+    NETWORK="-nic bridge,br=${bridge},model=virtio-net-pci"
+  else
+    # shellcheck disable=SC2054,SC2206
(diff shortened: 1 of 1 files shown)


diddledani (commented):
I don't understand how this is any different. The `args+=` that you've replaced is exactly the same as your inclusion via a `NETWORK` variable.

navycatt:
> I don't understand how this is any different. The `args+=` that you've replaced is exactly the same as your inclusion via a `NETWORK` variable.

Test it.

flexiondotorg:
What do you mean by "broken network"?

navycatt:
> What do you mean by "broken network"?

Look at issue https://github.com/wimpysworld/quickemu/issues/215.

navycatt:
It looks like the problem only occurs in Ubuntu Server. Since ubuntu server is not officially supported by quickemu, I close the PR.

PR #224, opened 2021-11-13
This adds OMVF support in solus

ZachBacon:
however secureboot in OMVF isn't enabled in solus, solus will have to enable secureboot in their omvf package before that can be enabled.


flexiondotorg:
Thanks!

PR #226, opened 2021-11-14
Fix case statement

lbschenkel:
The correct separator is "|" and not ",".


flexiondotorg:
Oh boy 🤦 Thank you for spotting this.

PR #235, opened 2021-11-16
Add Oracle Linux support

charles-hood:
Add support for Oracle Linux

### DIFF
--- a/quickget
+++ b/quickget
@@ -1046,11 +1046,17 @@ function get_oraclelinux() {
     local majorver=${RELEASE,1,1}
     local minorver=${RELEASE,3,1}
     
+    echo $RELEASE
+    echo $majorver
+    echo $minorver
+    
     local baseurl="https://yum.oracle.com/ISOS/OracleLinux/OL${majorver}/u${minorver}/${arch}/"
+    echo $baseurl
 #   local baseurl="https://download.rockylinux.org/pub/rocky/${RELEASE}/isos/${arch}"
 
     validate_release "releases_oraclelinux"
-    ISO="OracleLinux-R${majorver}-U${minorver}-${arch}-dvd.iso"
+    ISO="https://yum.oracle.com/ISOS/OracleLinux/OL8/u4/x86_64/OracleLinux-R8-U4-x86_64-dvd.iso"
+#    ISO="OracleLinux-R${majorver}-U${minorver}-${arch}-dvd.iso"
     URL="${baseurl}/${ISO}"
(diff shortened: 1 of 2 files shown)


philclifford:
You should probably add it to the Readme.md as well.

quickget
```diff
@@ -1023,6 +1038,30 @@ function get_opensuse() {
     make_vm_config "${ISO}"
 }
 
+function get_oraclelinux() {
+    local HASH=""
+    local ISO=""
+    local URL=""
+
+    local arch="x86_64"
```
philclifford, line 1046: I wonder if `arch="$(arch)"` would be useful

charles-hood:
> You should probably add it to the Readme.md as well.

Done. Added Rocky Linux to README as well, while I was in there adding Oracle Linux. It seemed logical to group them together since they are both based on sources from Red Hat Enterprise Linux.

quickget
charles-hood: > I wonder if `arch="$(arch)"` would be useful

Yes, perhaps, but my intention was to limit the scope of this first PR just to keep things simple. Is this something that could wait to be added later?

philclifford: could be - although it might be more trouble than help.  I'd  mentally file it for if/when the question arises in earnest.

quickget
```diff
@@ -1023,6 +1038,30 @@ function get_opensuse() {
...
+    
+    local majorver=${RELEASE::1}
+    local minorver=${RELEASE:2:1}
+  
+    local baseurl="https://yum.oracle.com/ISOS/OracleLinux/OL${majorver}/u${minorver}/${arch}/"
+    local hashurl="https://linux.oracle.com/security/gpg/checksum/OracleLinux-R${majorver}-U${minorver}-Server-x86_64.checksum"
+    
+    validate_release "releases_oraclelinux"
+
```
philclifford, line 1055: I think for 6.x and 7.x  its just  inserting "-Server" before "-${arch}" in the ISO (so a case/if with `major<8 then` logic and a local to hold the string (or not)
Mind you - I'd not bother with 6.x or 7.x myself, and be suprised if anyone wants <8.4 at this stage so the above isn't worth bothering with.  Obv. too much wittering so must be time for bed.

charles-hood: > I think for 6.x and 7.x its just inserting "-Server" before "-${arch}" in the ISO (so a case/if with `major<8 then` logic and a local to hold the string (or not) Mind you - I'd not bother with 6.x or 7.x myself, and be suprised if anyone wants <8.4 at this stage so the above isn't worth bothering with. Obv. too much wittering so must be time for bed.

Good catch! I added 7.9, 7.8 and 7.7 support. I agree it doesn't make sense to add 6.x at this point.

PR #243, opened 2021-11-19
Add Debian support

philclifford:
It is needed I feel.
This just finds the standard current releases (all DE options)

- [x] Will look to add the unofficial no-totally-free options as well.
  sorted similar to intel|nvidia for Pop! : standard or nonfree taken as options (or none)

nonfree not noted in the Readme as it doesn't mention intel/nvidia options either as far as I can see.
I think both need noting somewhere.

Could do with some refactoring around the DE options (like the windows languages but different, and applied here and to e.g. Garuda) but it is Friday night ....


philclifford:
[Heads-up](https://github.com/quickemu-project/quickemu-icons/issues/12) icon wizards

cowboygeneral:
Need to add Debian to function os_support()

daPhipz:
Icon PR done :heavy_check_mark:

Issue #162, opened 2021-10-27
Windows 10 and 11 crash every other boot


Lukewh:
#### Expected behaviour

Windows 10 and 11 boot ;)

#### Actual behaviour

Occasionally, the VM crashes at the 'BIOS' stage, and no Windows spinner displays.

#### Steps to reproduce the behaviour

Now I'm writing this report, it's not happening. I will endeavour to find a way to reproduce it consistently.

#### Quickemu output

```
~/vms quickemu --vm windows-11.conf 
Quickemu 2.2.7 using /usr/bin/qemu-system-x86_64 v6.0.0
 - Host:     Ubuntu 21.04 running Linux 5.11 (graeme-nolegs)
 - CPU:      Intel(R) Core(TM) i5-3470 CPU @ 3.20GHz
 - CPU VM:   1 Socket(s), 2 Core(s), 1 Thread(s), 8G RAM
 - BOOT:     EFI (Windows), OVMF (/usr/share/OVMF/OVMF_CODE_4M.fd), SecureBoot (off).
 - Disk:     windows-11/disk.qcow2 (64G)
 - MSR:      WARNING! Ignoring unhandled Model-Specific Registers is disabled.

             echo 1 | sudo tee /sys/module/kvm/parameters/ignore_msrs

             If you are unable to run macOS or Windows VMs then run the above 👆
             This will enable ignoring of unhandled MSRs until you reboot the host.
             You can make this change permenant by running: 'quickemu --ignore-msrs-always'
 - CD-ROM:   windows-11/virtio-win.iso
 - Display:  SDL, qxl-vga, GL (on), VirGL (off)
 - ssh:      On host:  ssh user@localhost -p 22220
 - SPICE:    On host:  spicy --title "windows-11" --port 5930 --spice-shared-dir /home/luke/Public
 - WebDAV:   On guest: dav://localhost:9843/
 - TPM:      windows-11/windows-11.swtpm-sock (36330)
 - Process:  Starting windows-11.conf as windows-11 (36332)

```

#### Linux Distribution & Kernel

Run the following and paste the output.

```bash
lsb_release --all
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 21.04
Release:	21.04
Codename:	hirsute

uname -a
Linux graeme-nolegs 5.11.0-38-generic #42-Ubuntu SMP Fri Sep 24 14:03:54 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

[...]

PR #244, opened 2021-11-20
Turn off Hibernation in Windows autounattend.xml

cowboygeneral:
Turn off Hibernation in Windows to correct issue #162
I've had some success correcting the "every other boot lockup problem" with Windows guests.
I was able to fix the problem on my machine by launching a cmd prompt with Admin privileges in the Windows guest and typing:
powercfg -h off 
(this turns off hibernation and thus also turns off fastboot)
This PR is an attempt to automate this fix using the autounattend.xml file.
Someone more familiar with Windows automation should test this thoroughly.
(There may be a better way to do it.)

### DIFF
--- a/quickget
+++ b/quickget
@@ -551,6 +551,11 @@ function unattended_windows() {
           <Description>Install spice-vdagent SPICE agent</Description>
           <Order>4</Order>
         </SynchronousCommand>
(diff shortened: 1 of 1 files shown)


flexiondotorg:
Thank you so much for contributing this much need fix! 🙂

PR #245, opened 2021-11-21
Add CachyOS support

vnepogodin:
Add support for CachyOS

### DIFF
--- a/quickget
+++ b/quickget
@@ -39,6 +39,7 @@ function pretty_name() {
     elementary)         PRETTY_NAME="elementary OS";;
     freebsd)            PRETTY_NAME="FreeBSD";;
     garuda)             PRETTY_NAME="Garuda Linux";;
+    cachyos)            PRETTY_NAME="CachyOS";;
     kdeneon)            PRETTY_NAME="KDE Neon";;
     linuxmint-cinnamon) PRETTY_NAME="Linux Mint Cinnamon";;
     linuxmint-mate)     PRETTY_NAME="Linux Mint MATE";;
@@ -126,6 +127,8 @@ function list_csv() {
         DOWNLOADER="${DL}"
       elif [ "${OS}" == "garuda" ]; then
         DOWNLOADER="${DL}"
+      elif [ "${OS}" == "cachyos" ]; then
+        DOWNLOADER="${DL}"
       elif [[ "${OS}" == *"kdeneon"* ]]; then
         DOWNLOADER="${DL}"
       else
@@ -155,6 +158,7 @@ function os_support() {
     freebsd \
     fedora \
     garuda \
+    cachyos \
     kali \
     kdeneon \
     kubuntu \
@@ -227,6 +231,10 @@ function releases_garuda() {
     xfce \
     mate \
     cinnamon
+}
+
+function releases_cachyos() {
+    echo 2021.11.10
 }
 
 function releases_kali() {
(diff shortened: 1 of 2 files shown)


quickget
```diff
@@ -810,6 +825,19 @@ function get_archlinux() {
...
 
+function get_cachyos() {
+    local HASH=""
+    local ISO=""
+    local URL=""
+    local VERSION=""
+
+    validate_release "releases_cachyos"
+    ISO="cachyos-cli-${RELEASE}-x86_64.iso"
+    URL="https://deac-fra.dl.sourceforge.net/project/cachyos-arch/cli-installer"
```
philclifford, line 836: I think this will not work:  you need to find a way to get and parse a mirror list or provide a static mirror URL  so that  the script can fetch the iso, ideally also a checksum file. If I run what the webget() will run here I get redirects and then an error. Also this script us used by a gui so needs to behave consistently for it.
And finally, given the aims of this project to be it would probably be advisable to include the gui-installer version as well as a release option.

philclifford (changes requested):
please try to find a better source with an associated checksum file

quickget
vnepogodin: Yeah there is no api for that yet
but I will try to make one

ptr1337: Hey @philclifford 

You can easily get the checksum from https://mirror.cachyos.org/ISO/cachyos-cli-2021.11.10-x86_64.iso
For example:  MD5=b944838bfea7ca9058ee599746f003d8d5330de813ce956dea2d9fd0d5a3545a

The next step is to create a router/loadbalancer for the mirros we have.

daPhipz:
As per the [About section](https://cachyos.org/?page_id=3752) of your webpage, I can see that you, @vnepogodin and @ptr1337, both are involved in the development of CachyOS. Do you have an SVG of the distro's logo, which we could use as an icon for a `.desktop` file that launches the VM?

daPhipz (changes requested):
I know that this might sound nit-picky and pedantic, but IMO it is still very important because of the huge amount of lines of code in `quickget`:
Please stick to the alphabetical ordering for every change you made in `quickget`, this will make it easier to find everything in the big file. Thanks! ;)

vnepogodin:
> As per the [About section](https://cachyos.org/?page_id=3752) of your webpage, I can see that you, @vnepogodin and @ptr1337, both are involved in the development of CachyOS. Do you have an SVG of the distro's logo, which we could use as an icon for a `.desktop` file that launches the VM?

Yes I have the logo https://cloud.ptr1337.dev/s/qgmpX5o87si9TND

daPhipz:
Thank you very much!

quickget
philclifford: Hi @ptr1337 if @vnepogodin updates the PR to set URL to "https://mirror.cachyos.org/ISO/" it would at least work for now.  The point of the checksum is for you to make them available (possibly with signature but we currently don't check that ) so that the script can get your ISO and you checksum and compare (because some users ask for this).  There is no  point whatever in me (or the script) downloading a file, computing a checksum then passing both to a function to see if they check out - I really hope they always will :slightly_smiling_face: ) .   So if you add a file (e.g. SHA256Sums )  with a list of checksums and filenames (or one for each file like ${file}.sha512sum) then the script can be made to work.  For some supported distros we even have to use a checksum file to find the sum and the filename (as it changes mostly daily at some unknown time...), but that works well and is there in the quickget source to re-use.

THope this helps.  Thanks

philclifford: s/THope/Hope/

philclifford (approved):
Installed OK. Thanks for making the requested changes. If a hash file turns up before this is merged or as a future PR so much the better, but this works. Thanks.

Issue #251, opened 2021-11-24
Document quickemu conf file


zenmaster24:
#### Expected behaviour

This is a question around usage - I'd like to know what options i can put in the conf file, and their syntax

#### Actual behaviour

I cant find this documented anywhere

#### Steps to reproduce the behaviour

`quickemu -h` outputs ` Usage quickemu --vm ubuntu.conf` which is not helpful

#### Quickemu output

```
$ quickemu --version
2.3.2
$ quickemu -h       

Usage
  quickemu --vm ubuntu.conf

You can also pass optional parameters
  --delete                : Delete the disk image.
  --display               : Select display backend. 'sdl' (default), 'gtk', 'none', or 'spice'
  --fullscreen            : Starts VM in full screen mode (Ctl+Alt+f to exit)
  --ignore-msrs-always    : Configure KVM to always ignore unhandled machine-specific registers
  --screen <screen>       : Use specified screen to determine the window size.
  --shortcut              : Create a desktop shortcut
  --snapshot apply <tag>  : Apply/restore a snapshot.
  --snapshot create <tag> : Create a snapshot.
  --snapshot delete <tag> : Delete a snapshot.
  --snapshot info         : Show disk/snapshot info.
  --status-quo            : Do not commit any changes to disk/snapshot.
  --version               : Print version
```

#### Linux Distribution & Kernel

```bash
$ lsb_release --all
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.3 LTS
Release:	20.04
Codename:	focal

$ uname -a
Linux $HOST 5.11.0-40-generic #44~20.04.2-Ubuntu SMP Tue Oct 26 18:07:44 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

philclifford:
Thanks for your feedback.

The script sets the following variable defaults and then reads the overrides that `quickget` or the user has written into the config file.

```
# Lowercase variables are used in the VM config file only
boot="efi"
bridge=""
cpu_cores=""
disk_img=""
disk_size=""
fixed_iso=""
floppy=""
guest_os="linux"
img=""
iso=""
macos_release=""
port_forwards=()
preallocation="off"
ram=""
secureboot="off"
tpm="off"
usb_devices=()
```
[...]

Issue #249, opened 2021-11-23
Add man pages for quickemu and quickget


caltrop2075:
#### Expected behaviour
a man page

Describe the your expected behaviour.

#### Actual behaviour
no man page

Describe the actual behaviour.

#### Steps to reproduce the behaviour

Outline the step to reproduce the issue you've encountered.

#### Quickemu output

Run `quickemu` or `quickget` and paste the output here.

#### Linux Distribution & Kernel

Run the following and paste the output.

```bash
lsb_release --all
uname -a
```

daPhipz:
This is to be expected, since nobody created a manpage for `Quickemu` yet.
I too think it would be a good idea to create one - feel free to start working on that, if you want! :)

PR #253, opened 2021-11-25
Building man pages

philclifford:
Edit: this has evolved somewhat and it is probably better to consider #334. Setting to draft for now rather than just closing.

I'm rubbish at man pages, but I slightly massacred the README into a couple of man pages for `quickget` and `quickemu`and a `quickemu_conf`one for the <vm>.conf parameters , via `pandoc` and a stolen makefile.
Then sliced into small pieces and cooked rapidly (see below)

Feel free to help or improve or replace entirely with something far better.

Fixes: #249 
Fixes: #251


philclifford:
if you have pandoc then run `make` to build local man pages, which you can view with `man -l <page>.1`

quickemu.1.md
```diff
@@ -134,57 +137,34 @@ See this (old) video where I explain some of my motivations for creating Quickem
...
-  * [util-linux](https://github.com/karelzak/util-linux)
-  * [sed](https://www.gnu.org/software/sed/)
-  * [spicy](https://gitlab.freedesktop.org/spice/spice-gtk)
-  * [swtpm](https://github.com/stefanberger/swtpm)
-  * [Wget](https://www.gnu.org/software/wget/)
-  * [xdg-user-dirs](https://www.freedesktop.org/wiki/Software/xdg-user-dirs/)
-  * [xrandr](https://gitlab.freedesktop.org/xorg/app/xrandr)
-  * [zsync](http://zsync.moria.org.uk/)
-
-# Install Quickemu
```
philclifford: If they have the man page surely they have already installed so not needed in the man page(s)

quickget.1.md
```diff
@@ -0,0 +1,227 @@
...
+
+quickget - download and prepare materials for building a quickemu VM
+
+# SYNOPSIS
+
+**quickget** [*os*] [*release*] [*variant*] | [*OPTION*]*
+
+# DESCRIPTION
+
+**quickget** will download the requisite materials and repare a configuration for `quickemu` to use to build and run
```
daPhipz, line 20:
```suggestion
**quickget** will download the requisite materials and prepare a configuration for `quickemu` to use to build and run
```
I spotted a typo here :)

daPhipz (changes requested):
Great work, I really like it! :+1:

Makefile
```diff
@@ -0,0 +1,37 @@
...
+	install -d $(DESTDIR)$(mandir)/man1
+	install -d $(DESTDIR)$(bindir)
+	#install -m 644 quickget.1 quickemu.1 quickemu_conf.1 $(DESTDIR)$(mandir)/man1
+	install -m 644 quickget.1 $(DESTDIR)$(mandir)/man1
+	install -m 644 quickemu.1 $(DESTDIR)$(mandir)/man1
+	install -m 644 quickemu_conf.1 $(DESTDIR)$(mandir)/man1
+	#install -m 644 quickgui.1 $(DESTDIR)$(mandir)/man1
+
+
+uninstall::
```
daPhipz, line 28: Why are there two `:`? Is this some Makefile magic I am too noob to understand, or a simple typo? XD
(There are two `:` in the commented out `clean` as well, in case this was an error)

philclifford: They are different. see [here](https://www.gnu.org/software/make/manual/html_node/Double_002dColon.html). As I am a noob who vaguely understands makefiles and spends a lot of time borrowing examples and then testing them until my improvements stop spraying files in all the wrong places I probably need to read the manual a few more times before I get the differences. So not a typo but possibly a different source of inspiration rather than actual intent.

philclifford:
Need to split the Readme and Man pages into "single version of truth" modular chunks, ideally with generated bits for supported OS and versions (from list_csv() ) so that they are consistent, somewhat self-maintaining and more automatable (and don't require general update /new distro PRs to touch Readme/docs . Makefile to use m4/pandoc to build main Readme.md and 3 (or more ) "man page" markdowns frrom parts, and then man pages from man page mds.

philclifford:
This is now superceded

PR #255, opened 2021-11-29
Initial gentoo support

guyosi:
This is initial support for Gentoo Linux to quickget,it does download and boot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ comprehensive support for macOS and Windows**.
   * [Ubuntu](https://ubuntu.com/desktop) and all the **[official Ubuntu flavours](https://ubuntu.com/download/flavours)**
   * [Fedora](https://getfedora.org/) & openSUSE ([Leap](https://get.opensuse.org/leap/), [Tumbleweed](https://get.opensuse.org/tumbleweed/), [MicroOS](https://microos.opensuse.org/))
   * [Linux Mint](https://linuxmint.com/) (Cinnamon, MATE, and XFCE), [elementary OS](https://elementary.io/), [Pop!_OS](https://pop.system76.com/)
-  * [Arch Linux](https://www.archlinux.org/), [Kali](https://www.kali.org/),[Garuda](https://garudalinux.org/), [ZorinOS](https://zorin.com/os/) & [NixOS](https://nixos.org/)
+  * [Arch Linux](https://www.archlinux.org/), [Kali](https://www.kali.org/),[Garuda](https://garudalinux.org/), [ZorinOS](https://zorin.com/os/), [NixOS](https://nixos.org/) & [Gentoo Linux](https://www.gentoo.org/)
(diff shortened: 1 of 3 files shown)


daPhipz:
When running `quickget gentoo latest`, I get

```
dirname: fehlender Operand
„dirname --help“ liefert weitere Informationen.
ERROR! Failed to download /20211128T170532Z/install-amd64-minimal-20211128T170532Z.iso. Try running 'quickget' again.
```

which roughly translates to

```
dirname: missing operand
„dirname --help“ for further information.
ERROR! Failed to download /20211128T170532Z/install-amd64-minimal-20211128T170532Z.iso. Try running 'quickget' again.
```

https://bouncer.gentoo.org/fetch/root/all/releases/amd64/autobuilds/${RELEASE}-iso.txt looks like this as of 2021-12-01 11:10 German time:

```
# Latest as of Wed, 01 Dec 2021 07:00:01 +0000
# ts=1638342001
20211128T170532Z/admincd-amd64-20211128T170532Z.iso 648019968
20211128T170532Z/install-amd64-minimal-20211128T170532Z.iso 457179136
```

My first guess is that `grep` fails, since there is no string `"Location"` anywhere in the file above.
Will investigate further!

philclifford:
> My first guess is that `grep` fails, since there is no string `"Location"` anywhere in the file above. Will investigate further!
It should find "Location" in the file it's `wget`ing.  Maybe you hit it during a build ??
It's worked reliably for me and the following still does
``` 
eval $( wget  -O/tmp/gentoolatest https://bouncer.gentoo.org/fetch/root/all/releases/amd64/autobuilds/${RELEASE}-iso.txt 2`>&1 |grep Location | awk '{print "LOCAT="$2}' )
    LOCAT=$(dirname ${LOCAT})
    eval $( awk '/admincd-amd64/ {print "ADMIN="$1}; /install-amd64-minimal/ {print "INSTALLMIN="$1}' /tmp/gentoolatest )
    URL="${LOCAT}/${INSTALLMIN}"
    ISO=$(basename "${INSTALLMIN}" )

```

Might need to quote the grep expression and also add something to catch absence (or oddness) in the `${RELEASE}-iso.txt `
And maybe quietly remove /tmp/gentoolatest at the end.

philclifford:
[Heads-up](https://github.com/quickemu-project/quickemu-icons/issues/12) icon wizards

philclifford:
@guyosi you'll need to do a little merging and conflict resolution to get this PR to be mergeable again (this is normal - we all step on each others toes at merge time ). Maybe worth setting this to draft until it is re-tested and confirmed working again once it has no conflicts.

@flexiondotorg note I may have added a bug to this that "works for me" but not @daPhipz  - see the [comment](https://github.com/quickemu-project/quickemu/pull/255#issuecomment-984193274) a couple above.

guyosi:
Finished squashing the bugs, there shouldn't be any conflicts.

PR #264, opened 2021-12-05
Re-disabled network for re-appearance of #115

### LINKED ISSUE #115: Windows 11 gets a corp/school iso...
When I get Win11 to boot it requires me to login to my work or school account?

philclifford:
Fixes #115 via re-imposition of #132

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,6 +294,13 @@ and [Windows 11](https://www.microsoft.com/en-gb/software-download/windows11)
 along with the [VirtIO drivers for Windows](https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/)
 and creates a virtual machine configuration.
 
+
+**During the Windows install network interfaces are completely disabled!** This
+is to allow those users who want to create local accounts to do so. A side affect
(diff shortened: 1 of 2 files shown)


diddledani:
I'd prefer to fix the issue instead of disabling networking - no network is a very bad user experience

philclifford:
> I'd prefer to fix the issue instead of disabling networking - no network is a very bad user experience

"as a user ...", apart from no longer being unable to install without a SAP SE login or their IT support, I cannot say I noticed. Was there an extra reboot in there ? I don't know.  Eventually getting to the desktop logged in and behold - network up, updates installed.  Obviously I'd also prefer a fix, but I can't help there .... but I look forward to this being closed in favour of something better.

flexiondotorg:
> I'd prefer to fix the issue instead of disabling networking - no network is a very bad user experience

I do agree, this is hammer of a workaround. So far, everything we've tested doesn't correct the situation and having SAP online accounts only is a far worse experience. So, I will add the workaround until we find something better.

PR #289, opened 2021-12-26
Add Haiku support

nqvrg:
Adding Haiku Beta 1 R3 support to quickget and quickemu.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ comprehensive support for macOS and Windows**.
   * [Oracle Linux](https://www.oracle.com/linux/) and [Rocky Linux](https://rockylinux.org/)
   * [Regolith Linux](https://regolith-linux.org/) (Release 1.6 and latest 2.0.0 pre-release  )
   * [FreeBSD](https://www.freebsd.org/) & [OpenBSD](https://www.openbsd.org/)
+  * [Haiku](https://www.haiku-os.org/)
   * Full SPICE support including host/guest clipboard sharing
   * VirtIO-webdavd file sharing for Linux and Windows guests
   * VirtIO-9p file sharing for Linux and macOS guests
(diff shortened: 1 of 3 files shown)


philclifford:
Does the project have a nice svg icon that can be used to build a [quickemu icon](https://github.com/quickemu-project/quickemu-icons/issues/12) ?

nqvrg:
Here they have their artwork, mostly png, but they do have the big ones in svg too: https://git.haiku-os.org/haiku/tree/data/artwork

Maybe the leaf over the letter H (or the H with the leaf) could be extracted and used for a quickemu icon. Example: https://git.haiku-os.org/haiku/plain/data/artwork/deskbar_leaf.png

Here are the usage rules of their trademark: https://www.haiku-inc.org/trademarks/allowed_usage/

nqvrg:
I just found this one: https://github.com/darealshinji/haiku-icons/blob/master/svg/App_Haiku3d.svg

I compared it to what I thought it was the original, and indeed, it looks like this person exported an svg version of the icon from an official one in their own icon format: https://git.haiku-os.org/haiku/tree/data/artwork/icons/App_Haiku3d

daPhipz:
Thanks for the links! I'll have a go at the icons in the near future.

daPhipz:
From https://www.haiku-inc.org/trademarks/allowed_usage/
> As part of a notice that your product incorporates [...] Haiku

Do you think this is the case with us?

nqvrg:
To be honest I am only a user of their system and have no contact with the developers, so I'm not qualified to advice about this.

However, I'd be surprised if we weren't allowed to have support for Haiku. This is after all one more way for it to be discovered, tested and used by users and programmers, and we're "using" their official ISOs, instead of redistributing them ourselves.

Maybe there is somebody involved in Quickemu who knows about trademarking and licenses could have a look and give a better confirmation?

philclifford:
I think it is concerning that the above link appears to prohibit our specific usage of their trademarks, and suggest we need to approach them for specific permission (and possible update of the above link).

PR #290, opened 2021-12-27
Update Debian version to 11.2.0

abe-101:
As i was **whimping** up a debian VM I Noticed it was at version 11.1.0. changed it to the latest release 11.2.0


abe-101:
I should probably note the i performed a test by building a Debian 11 VM with the latest Version of 11.2.0

PR #295, opened 2021-12-29
Replace multiple if statements with case and OR

the-mentor:
Pretty much the title says it all.
It simplifies the code and make it much shorter and easier to add new distros/OSs

Let me know if you have any questions

-DM

### DIFF
--- a/quickget
+++ b/quickget
@@ -510,7 +510,7 @@ function make_vm_config() {
 
     IMAGE_FILE="${1}"
     ISO_FILE="${2}"
-    case ${OS} in
+    case "${OS}" in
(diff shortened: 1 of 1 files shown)


quickget
```diff
@@ -510,77 +510,23 @@ function make_vm_config() {
...
-        IMAGE_TYPE="img"
-    elif [ "${OS}" == "windows" ]; then
-        GUEST="windows"
-        IMAGE_TYPE="iso"
-    elif [ "${OS}" == "zorin" ]; then
-        GUEST="linux"
-        IMAGE_TYPE="iso"
-    fi
-
+    case ${OS} in
```
diddledani, line 513:
```suggestion
    case "${OS}" in
```

diddledani (changes requested):
One fix, otherwise good.

quickget
the-mentor: Thanks for the review @diddledani.
Fixed!

PR #308, opened 2022-01-05
Add support for Tails Linux

sebcode:
Adds support for the stable release of Tails: https://tails.boum.org/

I had to add `boot="legacy"` otherwise it won't boot (neither ISO or IMG), not sure why.

### DIFF
--- a/quickget
+++ b/quickget
@@ -186,6 +186,7 @@ function os_support() {
     regolith \
     rockylinux \
     solus \
+    tails \
     ubuntu \
     ubuntu-budgie \
     ubuntu-kylin \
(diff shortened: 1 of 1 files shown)


quickget
```diff
@@ -55,6 +55,7 @@ function pretty_name() {
     popos)              PRETTY_NAME="Pop!_OS";;
     regolith)           PRETTY_NAME="Regolith Linux";;
     rockylinux)         PRETTY_NAME="Rocky Linux";;
+    tails)              PRETTY_NAME="Tails";;
```
philclifford, line 58: This line is not needed because it is done by the default case.

sebcode: You're right, I didn't see that. I removed that line.

PR #310, opened 2022-01-07
Add support for braille terminal

### LINKED ISSUE #307: add Braille terminal support
hello it will help me a lot if quickemu can add Braille display

references: 
* https://gitlab.archlinux.org/archlinux/archiso/-/blob/master/scripts/run_archiso.sh#L82
* https://lists.nongnu.org/archive/html/qemu-devel/2008-04/msg00052.html
* search braille in qemu manual

philclifford:
fixes #307

Added parameter `--acc | --accessible | --accessibility` which if given `braille` ( or `brl ` )
with add parameters to invoke `Qemu` braille terminal support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -521,6 +521,7 @@ Usage
   quickemu --vm ubuntu.conf
 
 You can also pass optional parameters
+  --accessible            : Select accessible facilitation. 'braille' (default - currently requires --display sdl )"
   --delete-disk           : Delete the disk image and EFI variables
   --delete-vm             : Delete the entire VM and it's configuration
   --display               : Select display backend. 'sdl' (default), 'gtk', 'none', or 'spice'
--- a/README.md
+++ b/README.md
@@ -370,6 +370,19 @@ SPICE ports the VM is connected to.
 ```bash
 cat ubuntu-focal/ubuntu-focal.ports
 ```
+# Accessibility
+
+Qemu provides support for using BrlAPI to display braille output on a real or fake device.
+
+
+```bash
+quickemu --vm ubuntu-impish.conf --accessibility braille
+```
+or even
+
+```bash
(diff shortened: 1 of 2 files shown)


alex19EP:
it doesn't work at the moment, but I think the problem is in qemu and not in the script. I'll try to figure it out.

philclifford:
> it doesn't work at the moment, but I think the problem is in qemu and not in the script. I'll try to figure it out.

Thanks - for me it seemed to write what I was expecting into the calling script, and qemu seemed to try and start the vm but failed because there was no braille display/device (in what seemed a sensible and correct way).  Is there any output you can share that might give a clue. Если логи на русском, это не проблема. Я немного читаю по русски. Плохо и очень медленно ...

I really hope you can figure it out and we can get this working.

alex19EP:
> Is there any output you can share that might give a clue.

i created [this](https://brltty.app/pipermail/brltty/2022-January/019079.html) thread on brltty's ml.

TLDR: all is working fign with sdl display but not with gtk one. don't know why yet.
so I think this can be merged.

p.s. мне всегда приятно видеть людей которые знают русский язык.

philclifford:
> TLDR: all is working fign with sdl display but not with gtk one. don't know why yet. so I think this can be merged.
Excellent news - we can always add tweaks (or extra documentation) later if there's a way to support other displays.  I'll add a note to the documentation

> p.s. мне всегда приятно видеть людей которые знают русский язык.
Я понимаю - и мне всегда приятно познакомиться с новым русским другом.
Thanks again for the request, and your help.

flexiondotorg:
Love this feature! Is there any reason why we should not always enable the braille character device(s) rather than have an option?

philclifford:
@flexiondotorg braille display feature breaks all but SDL display - until qemu is fixed it needs fencing off (see above)

PR #314, opened 2022-01-10
Use local isos

philclifford:
-optional parameters added
--localiso (if not supplied will use the distro/download filename
--isodir search root for find ...

copies in the first file found then carries on with wget/zsync

Also gathered the other parameters into the same fold.

Added usage() for --help (no parameters at all still behave as before since I think that is better UX)

### DIFF
--- a/quickget
+++ b/quickget
@@ -1585,6 +1585,38 @@ fi
 LANGS=()
 languages_windows
 
+# handle parameters
+
+# Take command line arguments
+#if [ $# -lt 1 ]; then
+#    usage
+#    exit 0
+#else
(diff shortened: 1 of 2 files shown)


quickget
```diff
@@ -1585,18 +1629,49 @@ fi
...
+# handle parameters
+
+# Take command line arguments
+#if [ $# -lt 1 ]; then
+#    usage
+#    exit 0
+#else
+    while [ $# -gt 0 ]; do
+        case "${1}" in
+          -delete|--delete|-delete-disk|--delete-disk)
```
philclifford, line 1641: Shellcheck points out this is not used ...

PR #321, opened 2022-01-16
add Void Linux support 

zen0bit:


daPhipz:
Thanks for your contribution!
Do you know if there is an icon for Void Linux (preferably in SVG format) that we could use in Quickgui?

zen0bit:
Icon merged...

quickget
```diff
@@ -352,6 +354,15 @@ function releases_ubuntu() {
     canary
 }
 
+function releases_void() {
+    echo i686-20210930 \
```
philclifford, line 358: I'd suggest not to include other releases while the get() only has one hard-coded option. Maybe also consider using the provided hash file to derive the latest iso for $RELEASE having given a choice of "cli or xfce" and optionally also "musl" flavour or not - examples exist to re-use (and I might send you a PR if time allows - no promises mind)

quickget
```diff
@@ -872,6 +892,23 @@ function get_openbsd() {
...
 }
 
+function get_void() {
+    local FLAVOR=""
+    local HASH=""
+    local ISO=""
+    local URL=""
+    local arch="x86_64"
+    # TODO:  only one hardcoded version of void added.. Add more and make it automatic
+    if
```
philclifford, line 902: `if` what ?  - has no test or `fi` - this will not work - please set to draft until it at least passes a simple smoke test:
```
./quickget void
./quickget: line 910: syntax error near unexpected token `}'
./quickget: line 910: `}'
```

philclifford (changes requested):
Please remove line 902 in `quickget` so this doesn't break it completely. Thanks.

quickget
zen0bit: repaired

philclifford (approved):
Thanks @zen0bit, much appreciated.

zen0bit:
I think could be merged...
check it

philclifford (approved):
LGTM
Thanks - they all pull nicely. A bit late for having a look at void but I'm looking forward to it when time permits.  Cheers.

philclifford:
In case it helps get them in I have merged into my consolidations of new distros and dealt with the conflicts that would have arisen. Here's hoping one way or another this will get in.  Thanks for the quick work today - it may prove timely.

zen0bit:
Just changed repository to original one..
Next step will be: Auto choose best mirror...

flexiondotorg:
Merged via https://github.com/quickemu-project/quickemu/pull/322

PR #322, opened 2022-01-17
Merging updated distros - Consolidated 15 PRs

philclifford:
Consolidated merge of  de-conflicted set of PRs adding or updating OS/releases and fixing release-specific bugs.

- #245
- #255
- #277 
- #281 
- #282 
- #284 
- #285 
- #286 
- #289 
- #290
- #292 
- #296 
- #298 
- #308 
- #321

PR #334, opened 2022-01-31
Build manpages and README.md plus consolidated new distros

philclifford:
Pulled the man page generation and partial README.md automation into alignment with the #322 PR
Includes a built README.md and man pages (and markdown versions) in ./docs

Fixes: #249 
Fixes: #251


flexiondotorg:
@philclifford Is this PR still relevant?

philclifford:
Subject to discussion and possible re-organisation, yes.  This has the ./docs with the man pages (man and markdown) , a Makefile to build the man pages from the markdown (and install man pages and/or "binaries"), plus a build directory with the raw "chunks" from which they are made up and the makefile to do that.

I think the end point might include getting the outputs (possibly with a makefile to install the man pages etc) into the "release" and the packaging but keep the "build" out of release (more for tidiness and size than anything - maybe a 
"submodule" approach would be the easiest from a workflow perspective and make it easiest to allow collaboration.  (There's a piece of work there too to make finding and organising the chunks easier but ...)

I'll drop the docs directory in another separate PR as getting man pages out would be good for the release.

philclifford:
closing pending discussions and decisions

Issue #339, opened 2022-02-04
quickget: feature suggestion: fzf


zen0bit:
#### Expected behaviour

use fzf

#### Actual behaviour

you must manually type choices

I think it would be nice can use fuzzy finder...

flexiondotorg:
I would prefer to not add additional dependencies to Quickemu

PR #343, opened 2022-02-06
Initial support for fzf

zen0bit:
Quick and dirty but its working for me..
Not all CDs tested...

#339

### DIFF
--- a/quickget
+++ b/quickget
@@ -1598,9 +1598,8 @@ if [ -n "${1}" ]; then
       exit 0
     fi
 else
-    echo "ERROR! You must specify an operating system:"
-    os_support
-    exit 1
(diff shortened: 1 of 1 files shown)


mark2185:
And what if the user doesn't have `fzf` installed?

zen0bit:
Could be closed solved in  #346 ...

flexiondotorg:
I agree with @mark2185 on this. Please add guard rails to catch the scenario where `fzf` is not available.

the-mentor:
@zen0bit i think you forgot to include  the releaselist file that you are catting.

Also i don't know if fzf is included in all the distros that quickemu supports

zen0bit:
repaired included in #353

PR #351, opened 2022-02-17
Netboot: update

zen0bit:


zen0bit:
Nice name chaged to netboot.xyz

[Leader ask for it...](https://github.com/netbootxyz/netboot.xyz/discussions/1074)

PR #353, opened 2022-02-17
Fzf: Added fzf support

zen0bit:


zen0bit (commented):
Work in Progress

help wanted

flexiondotorg:
Thank you for working on this, but I would prefer to not add additional dependencies to Quickemu. I do not see significant value in adding `fzf` support when compared to complicating the code.

This PR also used a needless `cat | grep`, something which should be avoided.

PR #354, opened 2022-02-17
Add colored output

zen0bit:


flexiondotorg (changes requested):
The variables `c1`, `c2` etc should be renamed to reflect the actual colour they represent, such as `$RED`, `$GREEN`.

Adding a dependency on `tput` is unnecessary, since `echo` can colourise without `tput`. 

In general, I am not a fan of arbitrary colours scattered in output. I would much prefer to see `fancy_message()` implemented so that colours are subtle and consistently applied:

  * https://github.com/wimpysworld/obs-fully-loaded/blob/main/obs-install.sh#L4

flexiondotorg:
Closing this as the current implementation is not suitable.

PR #355, opened 2022-02-17
Better specify release

zen0bit:


philclifford:
Awesome where it works.  Sadly doesn't for those like `ubuntu-mate` which have adopted the "OS-flavour" style.
If you split at the "-" it should work with the current patterns but I'd suggest we'd need to also change the documented guidance so contributors are clear to **only** use "-" in OS identifiers when adopting an "OS-flavour" approach and name the `releases_ <whatever>()` appropriately.

quickget
```diff
@@ -2133,70 +2133,8 @@ if [ -n "${2}" ]; then
...
-        releases_tails
-    elif [[ "${OS}" == *"ubuntu"* ]]; then
-        releases_ubuntu
-    elif [ "${OS}" == "void" ]; then
-        releases_void
-    elif [ "${OS}" == "windows" ]; then
-        releases_windows
-    elif [ "${OS}" == "zorin" ]; then
-        releases_zorin
+    if os_support | grep -q "${OS}" ; echo $? == 0 > /dev/null ; then
```
philclifford, line 2136: this test is not working:
```./quickget xxx
ERROR! You must specify a release: ./quickget: line 2137: releases_xxx: command not found
```
which is the same error if you try `quickget ubuntu-mate` or any of the "${OS}-*" types

please both fix the test, and account for all the "*" wildcards in the tests above.

philclifford (changes requested):

flexiondotorg:
Sorry, this pull request breaks quickemu.

Issue #363, opened 2022-02-19
Quickemu forces smartcard support, which breaks starting VMs on certain builds of qemu


laniusone:
#### Expected behaviour

The VM starts after running quickemu command, even without smartcard support. I find it non-essential feature to be there by default, since it causes problems.

#### Actual behaviour

The command ends with an error saying smartcard isn't supported.

#### Steps to reproduce the behaviour

Have qemu build without libcacard / smartcard support.

#### Quickemu output

```
Quickemu 3.11 using /usr/bin/qemu-system-x86_64 v6.2.0
 - Host:     Slackware 15.0 x86_64 running Linux 5.16 (voidstar.lanius.one)
 - CPU:      Intel(R) Core(TM) i5-9300H CPU @ 2.40GHz
 - CPU VM:   1 Socket(s), 2 Core(s), 2 Thread(s), 4G RAM
 - BOOT:     EFI (Linux), OVMF (/usr/share/OVMF/OVMF_CODE_4M.fd), SecureBoot (off).
 - Disk:     ubuntu-focal/disk.qcow2 (16G)
 - Display:  SDL, virtio-vga, GL (on), VirGL (on)
 - ssh:      On host:  ssh user@localhost -p 22220
 - SPICE:    On host:  spicy --title "ubuntu-focal" --port 5930 --spice-shared-dir /home/lanius/Public
 - WebDAV:   On guest: dav://localhost:9843/
 - 9P:       On guest: sudo mount -t 9p -o trans=virtio,version=9p2000.L,msize=104857600 Public-lanius ~/Public
 - smbd:     On guest: smb://10.0.2.4/qemu
qemu-system-x86_64: -chardev spicevmc,id=ccid,name=smartcard: unsupported type name: smartcard
allowed spice char type names: vdagent, usbredir
cat: ubuntu-focal/ubuntu-focal.pid: No such file or directory
 - Process:  Starting ubuntu-focal.conf as ubuntu-focal ()
```

#### Linux Distribution & Kernel

Slackware 15.0
Linux 5.16.9

PR #364, opened 2022-02-19
Issue#363 - make smartcard support optional

laniusone:
This fixes issue #363

### DIFF
--- a/README.md
+++ b/README.md
@@ -484,6 +484,13 @@ configuration:
 
   * `floppy="/path/to/floppy.img"`
 
+## Smartcard support
+
+If you want to be able to pass an SDCard to your VM, add the following line to
+your VM configuration:
+
+  * `smartcard=true`
+
 # File Sharing
 
 All File Sharing options will only expose `~/Public` (or localised variations)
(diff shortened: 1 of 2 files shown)


flexiondotorg (changes requested):
One of the design goals for Quickemu is to stand up as much as possible so it *"just works"*™ and limit to number of options/flags to keep it simple to use.

Instead of adding an option, it would be better if `quickemu` determines if smartcard support is available and enables it automatically when available.

laniusone:
@flexiondotorg Well, now quickemu has the floppy as an option, not default feature nor it checks if it is supported (which I believe would be always), that's why I suggested making it optional as well.

I might be wrong, but passing an SD Card is on par with floppy as a required feature + it breaks the script all together if smartcard is not supported. I might try to make it check if it's supported, but the problem is that even my qemu, which is compiled with smartcard support / libcacard, crashes because of the smartcard. So it might as well report having smartcard support, but still not work.

Similar issue exists in Gnome Boxes I believe, and the fix is just to not include smartcard by default.

laniusone:
It turns out my culprit was spice server built without smartcard support. The check for support or making it optional (since it might cause problems included by default) is still a good idea in my opinion, but I guess the best compromise here is to state in the docs what qemu and spice should support in order for the script to work. It would have saved me a lot of time.

PR #370, opened 2022-02-20
initial handy print url only feature

philclifford:
Prints out url in webget and exits.  

Trying to come up with some enhancements for future QA and testing as well as immediately useful info for troubleshooting.
Printing out the derived target download seemed an obvious starting point. 
TODO: 

- [ ] zsync and edge cases
- [ ] optional internal looping through all supported linux OS and release options
- [ ] optional link check (no download, just check the file is there/redirected to OK)

### DIFF
--- a/quickget
+++ b/quickget
@@ -573,6 +573,11 @@ function web_get() {
     local LOCALFILE=""
     local URL="${1}"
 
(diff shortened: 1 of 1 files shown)


philclifford:
not now possible

zen0bit:
why not posible?
i wanna this..

PR #374, opened 2022-02-21
Add FreeDOS support

nqvrg:
Adds support for FreeDOS versions 1.2 and 1.3 to quickget and quickemu.

There is one small issue I couldn't figure out how to solve yet:

The live session from version 1.3 works great, but when installing (both versions), after partitioning the disk, FreeDOS needs to reboot. Upon reboot, the ISO image is disabled and QEMU attempts to boot from disk, but the system is not yet installed.
I was able to complete the installation by booting QEMU directly from the command line, to make sure it works correctly, but I wasn't able to integrate it correctly into Quickemu.

I'll try to come up with a solution, but any suggestions would be welcome.

Worst case scenario, we could keep support for FreeDOS just as a live system, but it would be great to install it too.


flexiondotorg:
Perhaps preventing the ISO image from being unmounted when running FreeDOS will help? See this line: https://github.com/quickemu-project/quickemu/blob/master/quickemu#L620

nqvrg:
Thanks @flexiondotorg for your comment, I also thought about that.

However, Quickemu seems to be detecting correctly that the hard drive is not yet used. It says it's going to boot from the ISO, but QEMU doesn't seem to care and still tries to boot from disk.

**Example:**
- I run `quickemu --vm freedos-1.2.conf` for the first time.
- I partition the disk with the installer.
- I'm asked to reboot, I answer yes.
- QEMU tries to boot from hard drive.
- I close the QEMU window.
- I run `quickemu --vm freedos-1.2.conf` again.
- I see this message:
  ```
   - Disk:     freedos-1.2/disk.qcow2 (16G)
               Looks unused, booting from freedos-1.2/FD12CD.iso
   - Boot ISO: freedos-1.2/FD12CD.iso
  ```
- QEMU still tries to boot from hard disk.

I'm still not quite sure if I found a but or if I'm missing something... :thinking:

daPhipz:
Is there an SVG icon of some sorts that we could use for Quickgui?

nqvrg:
Here are the icons, @daPhipz

https://www.freedos.org/press-kit/logos.html
https://www.freedos.org/images/

nqvrg:
I think this is ready to go now. I hope it works correctly for everyone!

PR #379, opened 2022-02-22
Add some whitespace-formatting rules

nqvrg:
Every time I'm making changes to quickemu and quickget, I find myself indenting things with the wrong number of spaces, because the convention used in this project differs from my default.

I've also seen inconsistencies and therefore commits reformatting the code. So I added some Vim rules at the end of the files and an .editorconfig file, hopefully this will help!

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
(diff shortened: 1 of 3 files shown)


daPhipz:
Good idea!
For those who use VS Code and not vim, there's an extension that makes VS Code respect `.editorconfig` files: https://github.com/editorconfig/editorconfig-vscode

PR #385, opened 2022-02-23
Debian netinst nonsed

philclifford:
TIL how to do search replace in variables - never to old to learn new tricks.


nqvrg:
Wow! Very nice stuff, I didn't know this trick either. Thanks for sharing!!!

PR #403, opened 2022-02-27
Add batocera

philclifford:
by special request - awaiting confirmation from @popey ~~whether~~ but it now ~~boots~~ runs here.
Probably need user to run with spice and/or pass a controller through (because the UI is controller-centric! )
Confirmation or problems and suggestions welcome

### DIFF
--- a/quickemu
+++ b/quickemu
@@ -486,6 +486,15 @@ function vm_boot() {
         disk_size="16G"
       fi
       ;;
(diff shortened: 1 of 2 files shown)


flexiondotorg:
Please squash this so only the changes relevant to Batocera are included.

philclifford:
Hopefully that will now squash

Issue #382, opened 2022-02-22
Add Support for FreeDOS


pleappleappleap:
I'd like to be able to use this to spin up FreeDOS machines, with full video and sound.

philclifford:
Already a PR for FreeDOS  in hand #374

flexiondotorg:
@nqvrg After installing FreeDOS 1.2 or 1.3 the system is not able to boot.


![image](https://user-images.githubusercontent.com/304639/155620481-10f391bd-f5a5-4b68-a3f8-c10b3c10e776.png)

nqvrg:
This is annoying. I did several installations today. Most of the 1.2 ones were successful and ended up booting, and all of the 1.3 ones ended up booting correctly. I tried to boot now one of those machines I created earlier and I have the same problem as you.

I've tried changing some QEMU parameters (chipset, memory, disk size...) but no luck. I don't have enough knowledge about QEMU to find it.

I found on the FreeDOS wiki that VirtualBox is also affected: http://wiki.freedos.org/wiki/index.php/VirtualBox_-_Bugs:_JEMMEX

The workarounds they propose are rather bit esoteric, I haven't tried using the older version of `EMM386.EXE` `HIMEM.EXE`, but the "Safe" and "Emergency" modes do boot fine.

Any ideas about what could be adjusted on QEMU to fix this issue? There has to be a way...

chasecovello:
Removing "smm=off" from the -machine argument to qemu fixed this for me. You can edit the .sh script that quickemu creates, remove that option, and run the shell script directly, although if you run quickemu again it will overwrite any changes to that script.

PR #420, opened 2022-03-07
Fix freedos booting issue 

philclifford:
Solution to the booting problem discussed in #382 and elsewhere implementing the fix identified by @chasecovello  in https://github.com/quickemu-project/quickemu/issues/382#issuecomment-1059635250_


philclifford:
combined the two

PR #424, opened 2022-03-11
Added zero write detection to system disk configuration

chasecovello:
Writing zeroes to the free space on a virtual disk (to later shrink the image) causes the disk image file to expand to its maximum size. Adding "detect-zeroes=on" converts these to sparse writes for unallocated parts of the disk.


flexiondotorg (changes requested):
Please drop the commits regarding SOlaris support from this pull request

PR #428, opened 2022-03-22
Add initial Solaris/Illumos support

chasecovello:


philclifford:
Is it me or is this missing the needed additions to `quickget` ?

flexiondotorg (changes requested):
Please add the required additions to quickget so Solaris can be automatically downloaded.

PR #445, opened 2022-04-24
add install requirements to readme

abe-101:
Some package requirements have different names on different package managers.
To make it easier for new user we can include in the readme a copy-able one liner with all the packages needed for specific package managers.


abe-101:
It has been brought to my attention that there is the ppa for ubuntu and an aur package for arch that handle the dependency's.
But for other distros like Debian and Fedora this would be a nice addition.

Issue #452, opened 2022-05-02
Garuda is not working in quickget (releases and editions) not valid


dabrown645:
#### Expected behaviour

Expected garuda dr460nized gets downloaded

#### Actual behaviour

Describe the actual behaviour.

#### Steps to reproduce the behaviour

Outline the step to reproduce the issue you've encountered.

#### Quickemu output

bsh ⭆ quickget garuda 220131 dr460nized
ERROR! Failed to download http://mirrors.fossho.st/garuda/iso/garuda/dr460nized/220131/garuda-dr460nized-linux-zen-220131.iso with wget. Try running 'quickget' again.

After doing some research garuda nolonger supplies release 220131 release and in fact had 3 new releases in 3 days. I noticed that there was a latest directory on download site with all the current editions.

I have fixed an version of quickget and I am trying to figure out how to submit a pull request

#### Linux Distribution & Kernel

Run the following and paste the output.

```bash
lsb_release --all
uname -a
```

PR #453, opened 2022-05-02
Garuda update to quickget to support latest release

dabrown645:
* reformated quickget to use consistant 2 space indentation
* removed the vim config line so would respect .editorconfig
* updated .editorconfig to be consistant with quickemu
* updated garuda release function to support only latest release
* updated function editions_garuda to get valid editions from
  distribution web site
* updated function get_garuda to use the garuda latest iso images
* also change the default disk_size to 32G as I got error that 16G was
  to small when I tested

### DIFF
--- a/quickget
+++ b/quickget
@@ -290,23 +290,12 @@ function releases_freedos() {
 }
 
 function releases_garuda() {
-    echo 220131
+  echo latest
 }
 
 function editions_garuda() {
-    echo dr460nized \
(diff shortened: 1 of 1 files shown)


dabrown645:
I have backed out reformating quickget to consistent 2 space and the .editorconfig changes now it only does the garuda changes

philclifford:
Fixes #452

Issue #466, opened 2022-05-09
Can't start VM with Quickemu using  SDL display on Arch Linux


neuro-prog:
#### Expected behaviour

Start the VM without any issue

#### Actual behaviour

It won't start

#### Steps to reproduce the behaviour

Start a VM with the defaults

#### Quickemu output

Quickemu 3.15 using /usr/bin/qemu-system-x86_64 v7.0.0
 - Host:     "EndeavourOS Linux" running Linux 5.17 (eosbox)
 - CPU:      Intel(R) Core(TM) i7-4500U CPU @ 1.80GHz
 - CPU VM:   1 Socket(s), 1 Core(s), 2 Thread(s), 4G RAM
 - BOOT:     EFI (Linux), OVMF (/usr/share/OVMF/x64/OVMF_CODE.fd), SecureBoot (off).
 - Disk:     manjaro-gnome/disk.qcow2 (16G)
 - Display:  SDL, virtio-vga, GL (on), VirGL (on)
 - ssh:      On host:  ssh user@localhost -p 22220
 - SPICE:    On host:  spicy --title "manjaro-gnome" --port 5930 --spice-shared-dir /home/lorenzo/Pubblici
 - WebDAV:   On guest: dav://localhost:9843/
 - 9P:       On guest: sudo mount -t 9p -o trans=virtio,version=9p2000.L,msize=104857600 Public-lorenzo ~/Pubblici
qemu-system-x86_64: Display spice is incompatible with the GL context
cat: manjaro-gnome/manjaro-gnome.pid: No such file or directory

#### Linux Distribution & Kernel

```bash
LSB Version:    n/a
Distributor ID: EndeavourOS
Description:    EndeavourOS Linux
Release:        rolling
Codename:       rolling

Linux eosbox 5.17.5-arch1-1 #1 SMP PREEMPT Wed, 27 Apr 2022 20:56:11 +0000 x86_64 GNU/Linux
```

MrAureliusR:
Having the same issue. It seems perhaps the new version of qemu requires different arguments? I've tried with a bunch of my VMs that were all previously working, now none are working.

neuro-prog:
> **EDIT**: Temporary workaround: edit the vmname.sh file in the vmname/ directory. Remove `-device virtio-vga-gl,xres=1664,yres=936` and also remove `,gl=on` from the statement immediately afterward. The vm will launch, though it will not have GL acceleration.

[...]

PR #472, opened 2022-05-22
Add option to use spice-app as the --display

MatejSpindler:
All the plumbing to use spice-app as the viewer seems t be in place  already. 
First, commit adds spice-app to display_param_check().

The second commit reworks how the default viewer is selected. The variable display is never set, so the else part of the if statement can never be reached. 
Also, display_param_check() is already run when parsing command arguments.


MatejSpindler:
Forgot to mention, the default display sdl is crashing on qemu 7.0, could be an archlinux issue #466.

flexiondotorg:
Closed via #510

Issue #482, opened 2022-06-08
Add Linux Mint Debian Edition and update Mint to 20.3


tudo75:
Hi,
can you update Linux Mint to the latest version [20.3](https://linuxmint.com/download.php) and also add [Linux Mint Debian Edition](https://linuxmint.com/download_lmde.php)?

Thanks for all your work

#### Linux Distribution & Kernel

```bash
No LSB modules are available.
Distributor ID:	Linuxmint
Description:	Linux Mint 20.3
Release:	20.3
Codename:	una

Linux MINT-NEMO 5.4.0-113-generic #127-Ubuntu SMP Wed May 18 14:30:56 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

PR #484, opened 2022-06-08
add support for Linux Mint Debian Edition (LMDE)

philclifford:
requested in #482
will need another icon (or will it?) : heads-up https://github.com/quickemu-project/quickemu-icons/issues/12


daPhipz:
I don't think it will - does LMDE even have its own Icon? :thinking:

PR #488, opened 2022-06-15
Added Alma Linux 9.0 to the Alma Linux versions

infurno:
Added in Alma Linux 90 to the Alma Linux versions.


philclifford:
Already in #457 :-)

PR #501, opened 2022-07-08
Pop-OS 22.04 update

Michael-S:
Added 22.04, removed 21.10.
21.10 end-of-support is 2022-07-14.


Michael-S:
Ha! I see #493 already covers this, sorry.

PR #513, opened 2022-07-31
Oracle Linux 9.0 added

### LINKED ISSUE #514: Oracle Linux Release 9.0 needs adding
Oracle has removed references to some early 8.x releases too so the release list has been shortened and now reflects their published availability.

philclifford:
Some obsoleted 8.x releases also removed
Closes #514

### DIFF
--- a/quickget
+++ b/quickget
@@ -422,7 +422,7 @@ function releases_opensuse(){
 }
 
 function releases_oraclelinux() {
-    echo 7.7 7.8 7.9 8.2 8.3 8.4 8.5
+    echo 7.7 7.8 7.9 8.4 8.5 8.6 9.0
 }
 
 function releases_popos() {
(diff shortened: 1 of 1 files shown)


quickget
```diff
@@ -422,7 +422,7 @@ function releases_opensuse(){
 }
 
 function releases_oraclelinux() {
-    echo 7.7 7.8 7.9 8.2 8.3 8.4 8.5
+    echo 7.7 7.8 7.9 8.4 8.5 8.6 9.0
```
philclifford, line 425: 8.2 and 8.3 still exist and can be referenced but I think we should follow their lead since they no longer reference them on the download pages.

quickget
```diff
@@ -1270,8 +1270,8 @@ function get_oraclelinux() {
     local URL="https://yum.oracle.com/ISOS/OracleLinux/OL${VER_MAJ}/u${VER_MIN}/x86_64/"
 
     case ${VER_MAJ} in
-      8) ISO="OracleLinux-R${VER_MAJ}-U${VER_MIN}-x86_64-dvd.iso";;
-      *) ISO="OracleLinux-R${VER_MAJ}-U${VER_MIN}-Server-x86_64-dvd.iso";;
+      7) ISO="OracleLinux-R${VER_MAJ}-U${VER_MIN}-Server-x86_64-dvd.iso";;
```
philclifford, line 1273: As 9.x is consistent with 8.x it seemed sensible to swap the logic round.

PR #524, opened 2022-08-03
Add openindiana

philclifford:
Rebased #428 and added code to pull isos

This is held as draft because  out-of-the-box currently it kernel panics on booting (all isos, all my hosts) but can be made to try a little harder by setting boot and extra options to disable ACPI and force VESA.  However even with these in place the iso boots to a text console and goes into maintenance mode because the display manager fails to load.  The main culprit seems to be insisting on nvidia whether you have any or not

If anyone wants to try and beat into shape or take a look feel free .. They say spice is not supported but setting display to sdl or gtk got me no further.

(the installer credentials are jack/jack btw )

The installation guide says this re KVM : 
```
KVM | 
   OS type = Sun OpenSolaris
   Disk = SATA
   Remove USB Tablet
   NIC = e1000
   sound = AC97
   Processor = Copy host CPU configuration
   Disable CPU feature 'xsave'
   Video = QXL (QXL driver not supported, but the extra video memory helps)
   Display = VNC (Spice not supported)

```


philclifford:
@chasecovello feel free to take this back if you're still interested

philclifford:
It would be nice to, but needs help. 
I'm stuck so may as well be closed in the absence of Solaris fans with spare time.

PR #538, opened 2022-08-25
Remove duplicate section from README

mk3z:


philclifford:
Thanks for catching that! :shrug: no idea where the echo came from though ....
I've incorporated the fix in #540 (and hopefully made you co-author :crossed_fingers: ) so please close this one.

PR #547, opened 2022-09-11
fix for dynamically selecting debian releases

philclifford:
This should dynamically pick up debian releases for a while.
closes #546


flexiondotorg (changes requested):
Please rebase with git `master`

flexiondotorg (approved):

PR #566, opened 2022-10-07
Added check if file /etc/modprobe.d/kvm-quickemu.conf exists in function ignore_msrs_always()

tinsami1:
Function does nothing if file doesn't exists since grep fails.  Not sure if this is the intended behavior since the settings won't apply if the file does not exists.  So I added the file existence check condition.


flexiondotorg:
Thanks for the contribution, but this proposed change is not actually required 🙂

PR #569, opened 2022-10-17
Switch to disk devices that support TRIM

diddledani:
Currently, the virtio specification does not include provision for the TRIM (aka DISCARD) command that allows a guest operating system to signal the disk hardware that blocks have become unused so that the underlying device may clear the physical data.

The TRIM/DISCARD command was introduced for SSD disks as an extension to the AHCI specification that is used in SATA systems.

With Virtual Machines we can use this command to tell QEMU's Qcow2 driver to reclaim unused space in the disk image. This ensures the disk image file is kept to the smallest size possible where without the TRIM/DISCARD command it grows to it's maximum configured size and never shrinks again when data is deleted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,8 +351,6 @@ macos_release="catalina"
     particular macOS release.
     -   For example VirtIO Network and Memory Ballooning are available
         in Big Sur and newer, but not previous releases.
-    -   And VirtIO Block Media (disks) are supported/stable in Catalina
-        and newer.
 
 ### macOS compatibility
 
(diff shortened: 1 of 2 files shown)


diddledani:
I forgot to mention that `virtio-scsi-pci` is only _marginally_ less performant than `virtio-blk-pci` so I feel the tradeoff is worth it there for the gain of having disk images that don't *only* grow and *never* shrink again.

flexiondotorg (approved):

krackout:
Two points:
- On my installation, qemu 7.1, trim works on `virtio-blk-pci` also.
- macOS guests have been switched from `virtio-blk-pci` to `ahci` which may be considerably slower.

I revert to `virtio-blk-pci` for my VMs, I suggest that this change (from `virtio-blk-pci` to `virtio-scsi-pci` or `ahci`) should be reconsidered.

PR #579, opened 2022-11-02
adding ubuntu-unity 

interorbital:
This pull request adds Ubuntu Unity to quickemu and supports creating a VM with the 22.10 version, its first release as an official Ubuntu flavour.

However, I would like to provide a graceful way of downloading older versions so I'm submitting  this as a draft until I think of something.


interorbital:
Currently managed to display the [cdimage](http://cdimage.ubuntu.com/ubuntu-unity/releases/) download options for Ubuntu Unity. Added an if statement to clean up the `quickget` output.

I do not currently have the time to implement a way to download its releases pre-official flavour status but this is currently ready for review 😄

PR #606, opened 2022-12-24
Fix Pop!_OS install

goosepirate:
Pop!_OS 22.04 fails to install with 16G disk_size. With 32G it installs successfully.


flexiondotorg (approved):
Looks like Pop!_OS requires 20GB, so this looks like a sane default. Thanks.

PR #608, opened 2022-12-29
Added initial support for TrueNAS.

TenTypekMatus:
# Why?
Because there are many server Linux distributions and even some general-purpose Unix OSes. TrueNAS is an interesting server OS, where Core edition is based on FreeBSD and Scale version is based on Debian.


flexiondotorg (approved):
Thanks! LGTM

PR #616, opened 2023-01-21
Add Mageia

z3ntu:


quickget
```diff
@@ -1278,6 +1288,15 @@ function get_macos() {
     make_vm_config RecoveryImage.img
 }
 
+function get_mageia() {
+    local EDITION="${1:-}"
+    local HASH=""
+    local ISO="Mageia-${RELEASE}-Live-${EDITION}-x86_64.iso"
+    local URL="https://ftp.nluug.nl/pub/os/Linux/distr/mageia/iso/${RELEASE}/Mageia-${RELEASE}-Live-${EDITION}-x86_64/"
```
z3ntu, line 1295: Not sure what countries mirror is preferred, unfortunately there doesn't seem to be a CDN-like one that automatically chooses the closest. This one is from Netherlands.

philclifford: Thanks @z3ntu !

https://www.mageia.org/en/downloads/get/?q=Mageia-8-Live-Plasma-x86_64.iso tries to select an appropriate geographic mirror based on IP, so for example

```
wget -q https://www.mageia.org/en/downloads/get/?q=Mageia-8-Live-Plasma-x86_64.iso -O- |grep 'click here'|grep -o 'title=.*\.iso'|cut -d\" -f2
```
for me gets https://www.mirrorservice.org/sites/mageia.org/pub/mageia/iso/8/Mageia-8-Live-Plasma-x86_64/Mageia-8-Live-Plasma-x86_64.iso
so something like this might do the trick:

```
local ISO=$(wget -q https://www.mageia.org/en/downloads/get/?q="Mageia-${RELEASE}-Live-${EDITION}-x86_64.iso" -O- |grep 'click here'|grep -o 'title=.*\.iso'|cut -d\" -f2)
local HASH=$(wget  -q "${ISO}.sha512" -O- )
```

z3ntu: Sounds quite error print in case they change anything on their site, but if you think it's a good idea I can change it.

philclifford: I think it's a fair trade to take advantage of their mirrors against having to handle them possibly breaking it temporarily down the line.  They won't be the first or only (and the hard-coded mirror is almost as prone to their deciding to rename or restructure or a mirror going off-line) so I'd go for it.
Do please check my suggestion and improve it though - I was winging it and cooking lunch too so it will need QA :-) 

Entirely up to you - I can always chime in later anyway.
Thanks again for your contribution.

z3ntu: I've changed it to use href instead since that's what actually refers to the actual, I think it's more likely they change title than href

flexiondotorg (approved):
LGTM

PR #617, opened 2023-01-23
Correct basic error to get to downloading isos

philclifford:
This gets to download and run the installers with a 32G disk.
The 1st/2nd disk and boot disk nonsense in the conf gets ignored.
I'll  remove it when I get a moment.


philclifford:
missing AMD topext stuff ? Is that just this or is it generally oopsed ? ...

PR #659, opened 2023-03-02
Add Cereus linux

zen0bit:
Since I am void-linux fan...

 void-linux spin

https://cereus-linux.sourceforge.io

initial working commit

will need some refactoring?

Not tested shasums

### DIFF
--- a/quickget
+++ b/quickget
@@ -38,6 +38,7 @@ function pretty_name() {
     arcolinux)          PRETTY_NAME="Arco Linux";;
     cachyos)            PRETTY_NAME="CachyOS";;
     centos-stream)      PRETTY_NAME="CentOS Stream";;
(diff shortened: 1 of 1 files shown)


philclifford:
I think you'd improve this with a little refactoring : some of it subject to good guesswork or talking to the distro about their release plans, naming etc.  But maybe make the release 'beta' (and point that to .../beta/latest/... ) and using ${EDITION} in the path and target file.  Maybe better to use ${EDITION} (and maybe the ARCH ) to grep the iso from the page (or hash file) to minimise the need for PRs and updates if a new latest beta with a newer datestamp appears....

shasums seem to check out (with base anyway - if not then having provided the hash a fail would have been raised).

flexiondotorg:
I am not ad fan of adding endless forks of Void Linux.

PR #661, opened 2023-03-02
Add F-Void linux

zen0bit:
Since I am void-linux fan...

 void-linux spin

[F-void](https://sourceforge.net/projects/f-void)

initial working commit

will need some refactoring?


flexiondotorg:
I am not ad fan of adding endless forks of Void Linux.

PR #662, opened 2023-03-02
void-linux: all releases

zen0bit:
Since I am big void-linux fan...

 Added all released void-linux ISOs

will need some refactoring?

### DIFF
--- a/quickget
+++ b/quickget
@@ -599,11 +599,11 @@ function releases_ubuntu() {
 }
 
 function releases_void() {
-    echo current
+    echo current 20221001 20210316 20210930 20210218 20191109 20190526 20190217 20181111 20171007 20170825 20170220
 }
 
 function editions_void() {
-    echo glibc musl xfce-glibc xfce-musl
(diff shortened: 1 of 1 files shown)


quickget
```diff
@@ -599,11 +599,11 @@ function releases_ubuntu() {
 }
 
 function releases_void() {
-    echo current
+    echo current 20221001 20210316 20210930 20210218 20191109 20190526 20190217 20181111 20171007 20170825 20170220
```
philclifford, line 602: I think there's a reason they stopped shipping ISOs  with lots of DEs.  Personally I'd stop at current.

zen0bit: I have some Issues with latest ISO...
So i wanna keep track of all for testing purpose

up to you...

philclifford: Actually up to you ... Just that personally I'm uncomfortable encouraging "customers" to install 4- to 6-year-old isos ,even in VMs.  I mean you could continue to use this branch for your testing even if this is closed (and just keep it updated in your fork ...).  If there are issues that's an argument for including the 2022 and maybe 2021 but you're the void fan and I'm just guessing.. Certainly having recent releases without all the extra DE flavours and the resultant need to write a complex decision tree in bash would improve the chances of a merge .. but I'm just guessing ...

PR #663, opened 2023-03-05
Add VoidPup

zen0bit:
Since I am void-linux fan...

void-linux spin

[VoidPup](https://sourceforge.net/projects/vpup/)

[puppy discussion](https://forum.puppylinux.com/viewforum.php?f=180)

initial working commit

will need some refactoring?


zen0bit:
I wanna include all puppy releases but it will take time so

meanwhile...

flexiondotorg:

PR #665, opened 2023-03-06
Add SteamOS HoloISO

zen0bit:
HoloISO

SteamOS 3 (Holo) archiso configuration.

to your VM

[project](https://github.com/theVakhovskeIsTaken/holoiso)

Download is too slow...

will need some refactoring?

### DIFF
--- a/quickget
+++ b/quickget
@@ -64,6 +64,7 @@ function pretty_name() {
     reactos)            PRETTY_NAME="ReactOS";;
     rebornos)           PRETTY_NAME="RebornOS";;
     rockylinux)         PRETTY_NAME="Rocky Linux";;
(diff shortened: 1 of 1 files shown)


philclifford:
Ooof! I don't think you meant to merge that way round - this PR has a lot of others mixed in when it would be better just having https://github.com/quickemu-project/quickemu/pull/665/commits/2d5e839a62bc3e17d4401068d0e9c8768a25f0db

zen0bit:
Not intended

Back to HoloISO only

zen0bit:
Ok
what about PRETTYNAME?
Should i cut SteamOS also from there or not?

PR #669, opened 2023-03-07
Add AgarimOS

zen0bit:
You know abou my Void linux obsession?

I found other Void Linux fork...

[AgarimOS](https://sourceforge.net/projects/agarimos/)

So here it is..

Download
and Run
(Not all ISOs tested)


flexiondotorg:

PR #672, opened 2023-03-09
Add custom size for NixOS 

ahoneybun:
With the default a fresh install has 3% disk space left which isn't much to work with and even worse once updates are applied.


philclifford (commented):
Why not put in into the 32G case at the top?  16 was clearly too small but adding an in-between case seems unnecessary, particularly for an OS with so much software to try as well as the inherent elbow room trying nix may need.

ahoneybun:
@philclifford I didn't really have a reason for that location, I've moved it to the top with the 32G case as that would work as well.

PR #696, opened 2023-04-27
Added discard support and zero write detection to system disk configuration

chasecovello:
I created a PR for this a while back, but it had other irrelevant commits and I never fixed it, so opening a new one with just this commit.

Adding the "discard=unmap" option to the system disk causes qemu to unmap (i.e., convert into sparse extents) regions of the disk image identified by the guest's use of the trim/discard command.

Writing zeroes to the free space on a virtual disk (to later shrink the image) causes the disk image file to expand to its maximum size. Adding "detect-zeroes=on" converts these to sparse writes for unallocated parts of the disk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,7 +326,8 @@ supported.
 -   Use cursor keys and enter key to select the **macOS Base System**
 -   From **macOS Utilities**
     -   Click **Disk Utility** and **Continue**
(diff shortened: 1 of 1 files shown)


flexiondotorg:
We've implemented and reverted similar changes in the past, this change will require confirmation that it still works with Windows 10/11 and supported macOS versions.

PR #697, opened 2023-04-27
Add initial Solaris/Illumos/OpenIndiana support

chasecovello:


philclifford:
Thanks for coming back to this @chasecovello .  

Am I being dim?  The gui iso booted fine, installed apparently OK  with the gui installer but the resulting virtual disk doesn't boot for me.  Trying again with the text installer but is there something that needs doing after installing and before rebooting (that I need to add to the docs for all the curious folk who will fire it up, hit the defaults and end up only rebooting into the iso installer  ?

chasecovello:
I didn't catch this earlier because I only used the text mode installer, not GUI. I think I figured it out. The installer defaults to creating an MBR partition table without an EFI system partition. The text installer correctly creates a GPT table with EFI partition only if you select to install on the "whole disk" and not in a partition. The GUI installer appears to have a bug that creates an MBR regardless of which option you choose.

However, all four installation methods (GUI/whole disk, GUI/partition, text/whole disk, text/partition) work with BIOS boot. Text/whole disk still creates EFI/GPT, but makes it bootable with BIOS as well. So I've modified quickget to write 'boot="legacy"' to the config file. After doing test installs with all four methods, it seems to work. I've pushed the changes to my repo (along with a USB controller fix -- ehci seems to lose interrupts, xhci works).

flexiondotorg (approved):
LGTM

Issue #591, opened 2022-11-23
Request: line breaks in generated script


blobfish:
Thank you.

It is a pain to look at the generated qemu script when testing different options outside quickemu. Mine is 1 line of 2450 characters. Word wrap gets you by, but it would be nice to have some escaped line breaks for some human formatting.

PR #707, opened 2023-05-01
Issue591 - line breaks in generated script

dabrown645:
I had a typo in first pull request sorry


philclifford:
If you add
Closes #591
in a line on its own to your first comment, then if/when this is merged the issue will be closed too, keeping things neat and saving maintainers' time. Also it helps everyone because the link appears at both ends over there  :point_right:

PR #708, opened 2023-05-03
fix url and iso name for rockylinux versions

Wojberni:
Hi, I saw an issue for rockylinux images. The images that the quickget was pulling had wrong URL, so I fixed it. Hopefully everything is fine and you will be able to merge it.


philclifford (approved):
LGTM

Issue #709, opened 2023-05-04
Feature request: add Icaros Desktop or other AROS distributions


eylenburg:
It would be interesting to [AROS](https://aros.sourceforge.io/) supported in Quickemu. AROS is a free & open source reimplementation of Amiga OS that runs on x86 computers that's been in development since the mid-90s. There are several AROS distributions, the most well-known appears to be [Icaros Desktop](https://vmwaros.blogspot.com/) but there are others such as [AROS One](https://sites.google.com/view/arosone).

flexiondotorg:
Anyone interested in working on this might find our [guide to adding a new OS to quickget](https://github.com/quickemu-project/quickemu/wiki/06-Advanced-quickget-features#adding-a-new-os-to-quickget) useful.

shmahabi-dev:
This issue is being closed due to inactivity for 814 days. If this is still relevant, please comment or reopen and it will be re-opened. Thanks!

shmahabi-dev:
This issue is being closed due to inactivity for 814 days. If this is still relevant, please comment or reopen and it will be re-opened. Thanks!

eylenburg:
still think it would be cool to have

PR #712, opened 2023-05-09
Add AROS One

philclifford:
Seems to be working, needs testing and possibly improving some more and documenting how to get through a basic installation for the curious.

Partly addresses #709

Getting isos from Google Drive was a challenge !


flexiondotorg:
Please file a new PR if you still want to add AROS One.

PR #713, opened 2023-05-10
Add icaros desktop

philclifford:
Alternative AROS distro.  Fetches and prepares VM but booting the iso
tends to imitate Guru Meditation or black screen of limbo or boot loops. Will need more effort to get to a running live system.

The other distro mentioned in #709. At the moment it looks like Aros One (#712 ) will make it and this one will not ...


flexiondotorg:
Please file a new PR if you still want to add icaros.

PR #752, opened 2023-09-14
Add Miyo Linux

zen0bit:


zen0bit:
Here is icon
[miyo](https://sourceforge.net/p/miyolinux/discussion/general/thread/2a9b8d1816/5a88/c7a5/attachment/37273205.png)

quickget
```diff
@@ -1503,6 +1513,25 @@ function get_manjaro() {
     echo "${URL} ${HASH}"
 }
 
+function get_miyo() {
+    local EDITION="${1:-}"
+    local HASH=""
+    local DATE=""
+    local ISO="miyo-${EDITION}-x86_64-BIOS-${DATE}.iso"
```
philclifford, line 1520: AS DATE is unset this will be broken and is also overwritten in the `case` below so please remove the ISO assignment and the unused DATE (or assign and use it in the EDITION `case` options )

philclifford (changes requested):

philclifford (approved):

philclifford:
> Here is icon [miyo](https://sourceforge.net/p/miyolinux/discussion/general/thread/2a9b8d1816/5a88/c7a5/attachment/37273205.png)
If there were an SVG it would be better for the [art department](https://github.com/quickemu-project/quickemu-icons/issues/12)

zen0bit:
Not too familiar with Inkscape, but here it is...
[miyo.svg](https://raw.githubusercontent.com/quickemu-project/quickemu-icons/d8b659566f437293f96aa6de6a0a3dc8e5fd7566/src/distro-icons/miyo.svg)

flexiondotorg:
I am going to politely decline this one. It is a fringe Devuan respin. I think we need to create some guidelines for what distros will be added, because quickget becoming "install distrowatch" is not going to be sustainable.

PR #756, opened 2023-09-15
Add whonix

### LINKED ISSUE #740: Whonix Support?
Can we get Whonix support? I think that would be good because we have tails so whonix would also be good. This would also be good because Whonix takes some more steps then a one click install with a iso.

zen0bit:
Closes #740

Since whonix comes in 2 gcow2 images process slightly differ from others supported OSes (Which comes on ISOs)

- [x] Get images
- [x] Hash check
- [x] Upack images
- [x] Read license agreement
- [x] Accept/Decline license agreement
- [ ] Setup
- [ ] Run Whonix gateway
- [ ] Run Whonix workstation

Help appreciated

### DIFF
--- a/quickget
+++ b/quickget
@@ -274,6 +274,7 @@ function os_support() {
     vanillaos \
     void \
     vxlinux \
+    whonix \
     windows \
     xerolinux \
     xubuntu \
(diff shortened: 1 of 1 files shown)


zen0bit:
#767

zen0bit:
whonix

lj3954:
I've made some progress on this. Both VMs are now launched (output of the second VM should probably be piped to /dev/null). Internet connectivity must be fixed, though. 

https://github.com/lj3954/quickemu/tree/whonix-test

ehula:
Whatever happened to this? It would be great to have.

PR #765, opened 2023-09-16
Fix Gentoo hash check in quickget

Neinei0k:
Gentoo DIGESTS file contains BLAKE2B and SHA512 hashes. Since BLAKE2B comes first in the file, it was incorrectly selected when checking SHA512 checksum.


zen0bit (approved):
Works for me

PR #782, opened 2023-10-02
get rid of linux in short names

zen0bit:


flexiondotorg:
Going to reject this; we shouldn't change the well-known distro names the upstream projects use.

zen0bit:
But this don't change naming.

Just short names used for downloadïng/running VMs.

Pretty names stays same' so users will see fullname anyway.

Just less typing...

PR #795, opened 2023-10-11
Add Zephix /no hash/

zen0bit:
No hash yet provided.

Contacted author...
zephix


flexiondotorg:
A live-only distribution. I'm not compelled to add this to quickemu.

PR #826, opened 2023-10-27
resizing screen support for wayland like available for x11

dabrown645:
Added support to resize screen like available for x11 - also reformatted section to standard 4 space indentation

### DIFF
--- a/quickemu
+++ b/quickemu
@@ -683,41 +683,41 @@ function vm_boot() {
     *) AUDIO_DEV="pa,id=audio0";;
   esac
 
-  # Determine a sane resolution for Linux guests.
-  if [ "${guest_os}" == "linux" ]; then
-    local X_RES=1152
-    local Y_RES=648
-    if [ "${XDG_SESSION_TYPE}" == "x11" ]; then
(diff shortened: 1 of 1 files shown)


TuxVinyards:
I like this screen resizing work that you are doing.  I have been able to do some neat things with it in qqX. 

I haven't tested your Wayland changes but I am going to assume that they work. 👍

However, can't say that I like or agree with you on 4 space tabbing.  I think that most people are moving towards 2 now.

It's a bit strange having part of the quickemu project on 2 space & part of it on 4, I know.  I have taken to adding vim tab statements to my code now.  I found them in @flexiondotorg 's code.  I don't know if he uses vim but basic text editors seem to pick them up and it seems to work well.

For me, you can get much more on the page with 2 space and it's much easier to read now that we have more modern and better highlighting IDE's  ...

dabrown645:
I agree with you that 2 is what I prefer. I opened an issue about consistency and was waiting on decision on 2 or 4 then I noticed that someone posted a pull request for quickget (817) with indent of 4 so I submitted quickemu with 4 for consistency. I all agree I am willing change to 2 but in the meantime consistency is the best choice.

PR #847, opened 2023-11-19
Add options to directly specify VM width and height on cmdline or config

doqfgc:
This PR adds `--width` and `--height` arguments, as well as `width` and `height` config options, overriding hardcoded resolutions when specified.

It looks like this may only apply to Linux VMs, but it may prove useful in the event that automatically resizing fails.


flexiondotorg:
Closed via #986

Issue #853, opened 2023-11-25
Fedora does not list mirrors for ISOs of unsupported versions (currently pre-38)


lj3954:
https://fedoraproject.org/releases.json

Quickget breaks while downloading any fedora ISO for a version not listed here. The supported releases goes all the way back to 33.

PR #854, opened 2023-11-26
Fix quickget creating empty directories while using show-iso-url or test-iso-url parameters

lj3954:
Quickget currently creates a directory before checking the values of the show-iso-url or test-iso-url variables. If either are called, directory is left empty and serves no purpose.


lj3954:
Also fixes #853

flexiondotorg (approved):
LGTM

PR #855, opened 2023-11-26
Beautify output and make it more consistent between operating systems. 

lj3954:
non-macOS - print out "Downloading (OS) (Release) (Edition, if applicable) from (URL)" prior to download.

Windows - added show-iso-url and test-iso-url functionality

macOS - added show-iso-url and primitive test-iso-url, although they serve little purpose due to Apple restricting the download based on HTML headers (previous behavior: only print out/check link for OpenCore.qcow2 file)
                Slightly change output ("Downloading ventura" > "Downloading macOS ventura", added "Downloading macOS firmware"


lj3954:
Included in #860. May cause conflicts with it. Closing

PR #860, opened 2023-11-29
Replace macrecovery with inbuilt downloading and alternative method of verification

lj3954:
- Fetch URLs and HTML headers from Apple. Essentially just a recreation of some of the functions used in python macrecovery.
- Implement HTML headers in web_get method, in order to enable the downloading of the macOS image and checksum files. macOS images can be downloaded using wget or aria2, exactly like other files. 
- Use the python version of 'chunkcheck' (https://gist.github.com/MCJack123/943eaca762730ca4b7ae460b731b68e7) which was written by zhangyoufu, in order to verify the image using Apple's chunklist file. This is exactly the same code macrecovery uses to verify the image, I just added the main function so that it can compile. Unfortunately, it appears to be unfeasible to read this chunklist file and verify the image against its sha256sums within the shell script.  
- Slightly changed output for the download of all operating systems, #855. 
- show-iso-url and test-iso-url are now fully functional for macOS and Windows.
- macOS Sonoma added


lj3954:
One potential change that could be made is implementing HTML headers in the web-get method. Since it would only be used for macOS image downloading, I decided to keep a separate function with the specific headers necessary instead. I'm not particularly familiar with aria2, and reading the documentation it looks like specifying HTML headers would be quite difficult, so I'm not implementing that in this PR.

lj3954:
Fixes #865

lj3954:
wget2 cannot download the chunklist, similarly to aria2. I'm not sure exactly why, perhaps due to parallel downloading. cURL works correctly for this and will be used instead (5824652). Wget, wget2, and aria2 all are capable of downloading the macOS dmg file correctly.

PR #862, opened 2023-11-30
remove empty lines inside functions /NO CODE CHANGE/

zen0bit:


flexiondotorg:
I am not inclined to accept this change as it will degrade readability IMO. Can you provide some justification to convince me otherwise?

PR #871, opened 2023-12-14
add qrun 0.1 - Simple VMs runner for quickemu

zen0bit:
Use gum instead of fzf

Feature complete

But #864 needs to be merged for tip to work


zen0bit:
New polishing update

PR #879, opened 2023-12-18
Fix wget2 with quickget (#878)

### LINKED ISSUE #878: Downloading fails when wget2 is used to replace wget
Fedora (https://fedoraproject.org/wiki/Changes/Wget2asWget) and other distros are starting to replace wget with wget2. Wget2 renames the "show-progress" option to "force-progress", causing downloads from quickget to fail.

lj3954:
Wget2 renames "show-progress" to "force-progress". This fixes the command when wget2 is installed. 
Fixes #878

### DIFF
--- a/quickget
+++ b/quickget
@@ -1253,7 +1253,7 @@ function get_archcraft() {
     local URL=""
     local TMPURL=""
 
-    TMPURL=$(wget -q -S -O- --max-redirect=0 "https://sourceforge.net/projects/archcraft/files/latest/download" 2>&1 | grep -i Location | cut -d' ' -f4)
+    TMPURL=$(curl -Lfs "https://sourceforge.net/projects/archcraft/files/latest/download" -w %{url_effective} -o /this/is/a/nonexistent/directory/$RANDOM/$RANDOM)
     URL=${TMPURL%\?*}
     echo "${URL} ${HASH}"
 }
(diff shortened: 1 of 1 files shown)


quickget
```diff
@@ -939,14 +939,20 @@ function web_get() {
       exit 1
     fi
 
+    if wget --help | grep -q Wget2; then
```
zen0bit, line 942: if wget -V | grep -q Wget2; then

lj3954: I think it's actually better to check whether wget2 is installed in the first place and use it if so; some distributions may not ship this package that symlinks /usr/bin/wget to /usr/bin/wget2. Wget2 is faster and should be used instead whenever available.

lj3954: Also, just remembered, the best way to check whether /usr/bin/wget leads to wget2 would be to use `readlink /usr/bin/wget` and check whether the output is equal to 'wget2'. That would be much faster than any other way, it only takes 2 milliseconds on my system compared to 6 with wget2's help and version menus.

lj3954:
Wget2's output also does not work correctly for finding where a URL redirects. Instead, use curl and direct output to a nonexistent directory in order to only receive the final URL (/dev/null wouldn't work, since it's valid and you can download a file to it).

24fpsDaVinci:
thank you, this patch worked. maybe there could be a wget2 branch?

PR #889, opened 2023-12-29
Add Slint

zen0bit:
Slint is an easy-to-use, versatile, blind-friendly Linux distribution for 64-bit computers. Slint is based on Slackware and borrows tools from Salix. Maintainer: Didier Spaier.


zen0bit:
continues in #892

philclifford (commented):
LGTM
please re-open this for consideration

Issue #717, opened 2023-06-06
qemu-system-x86_64: -device ccid-card-passthru,chardev=ccid: 'ccid-card-passthru' is not a valid device model name


neevek:
#### Expected behaviour

```
Successfully starting the VM.
```

#### Actual behaviour

```
Failed
```

#### Steps to reproduce the behaviour

```
sudo quickemu --vm ubuntu-22.04.conf --display none
```

#### Quickemu output

```
Quickemu 4.7 using /usr/local/bin/qemu-system-x86_64 v8.0.2
 - Host:     Debian GNU/Linux 11 (bullseye) running Linux 5.10 (debian)
 - CPU:      AMD Ryzen 7 7735HS with Radeon Graphics
 - CPU VM:   1 Socket(s), 4 Core(s), 2 Thread(s), 8G RAM
 - BOOT:     EFI (Linux), OVMF (/usr/share/OVMF/OVMF_CODE.fd), SecureBoot (off).
 - Disk:     ubuntu-22.04/disk.qcow2 (16G)
             Looks unused, booting from ubuntu-22.04/ubuntu-22.04.2-desktop-amd64.iso
 - Boot ISO: ubuntu-22.04/ubuntu-22.04.2-desktop-amd64.iso
 - Display:  NONE, virtio-gpu, GL (on), VirGL (off)
 - ssh:      On host:  ssh user@localhost -p 22220
 - SPICE:    On host:  spicy --title "ubuntu-22.04" --port 5930 --spice-shared-dir /root
 - WebDAV:   On guest: dav://localhost:9843/
 - 9P:       On guest: sudo mount -t 9p -o trans=virtio,version=9p2000.L,msize=104857600 Public-root ~/root
 - Network:  User (virtio-net)
 - Monitor:  On host:  nc -U "ubuntu-22.04/ubuntu-22.04-monitor.socket"
             or     :  socat -,echo=0,icanon=0 unix-connect:ubuntu-22.04/ubuntu-22.04-monitor.socket
 - Serial:   On host:  nc -U "ubuntu-22.04/ubuntu-22.04-serial.socket"
             or     :  socat -,echo=0,icanon=0 unix-connect:ubuntu-22.04/ubuntu-22.04-serial.socket
qemu-system-x86_64: -device ccid-card-passthru,chardev=ccid: 'ccid-card-passthru' is not a valid device model name
cat: ubuntu-22.04/ubuntu-22.04.pid: No such file or directory
[...]

PR #890, opened 2023-12-29
Check whether smartcard is supported

### LINKED ISSUE #888: bug: VM fails to start because smartcard support is absent
#### Expected behaviour

quickemu --vm <distro> <version> should start the VM

#### Actual behaviour

The VM fails to start.

 Please note that I have installed qemu 8.2.0. [Configuration summary](https://slackware.uk/slint/x86_64/slint-15.0/source/qemu/LOG_8.2.0). 

#### Steps to reproduce the behaviour

Outline the step to reproduce the issue you've encountered.

```
quickget windows 11
quickemu --vm windows-11.conf
```
I have tried other distributions, same issue in all cases.
#### Quickemu output

Run `quickemu` or `quickget` and paste the output here.

```
quickemu --vm windows-11.conf
Quickemu 4.9.1 using /usr/bin/qemu-system-x86_64 v8.2.0
 - Host:     Slint 15.0 running Linux 6.6 (ici.ici.net)
 - CPU:       Intel(R) Core(TM) i7-4770 CPU @ 3.40GHz
 - CPU VM:   1 Socket(s), 2 Core(s), 2 Thread(s), 4G RAM
 - MSR:      WARNING! Ignoring unhandled Model-Specific Registers is disabled.

             echo 1 | sudo tee /sys/module/kvm/parameters/ignore_msrs

             If you are unable to run macOS or Windows VMs then run the above 👆
             This will enable ignoring of unhandled MSRs until you reboot the host.
             You can make this change permanent by running: 'quickemu --ignore-msrs-always'
 - BOOT:     EFI (Windows), OVMF (/usr/share/qemu/edk2-x86_64-code.fd), SecureBoot (off).
 - Disk:     windows-11/disk.qcow2 (64G)
             Looks unused, booting from windows-11/Win11_23H2_English_x64v2.iso
 - Boot ISO: windows-11/Win11_23H2_English_x64v2.iso
 - CD-ROM:   windows-11/virtio-win.iso
 - Display:  SDL, virtio-vga-gl, GL (on), VirGL (on)
 - Sound:    intel-hda
 - ssh:      On host:  ssh user@localhost -p 22220
 - WebDAV:   On guest: dav://localhost:9843/
 - smbd:     On guest: smb://10.0.2.4/qemu
 - TPM:      windows-11/windows-11.swtpm-sock (6844)
 - Network:  User (virtio-net)
 - Monitor:  On host:  nc -U "windows-11/windows-11-monitor.socket"
             or     :  socat -,echo=0,icanon=0 unix-connect:windows-11/windows-11-monitor.socket
 - Serial:   On host:  nc -U "windows-11/windows-11-serial.socket"
             or     :  socat -,echo=0,icanon=0 unix-connect:windows-11/windows-11-serial.socket
qemu-system-x86_64: -no-hpet: warning: -no-hpet is deprecated, use '-machine hpet=off' instead
qemu-system-x86_64: -chardev spicevmc,id=ccid,name=smartcard: unsupported type name: smartcard
allowed spice char type names: vdagent, usbredir
cat: windows-11/windows-11.pid: No such file or directory
 - Process:  Starting windows-11.conf as windows-11 ()
 - Sending:  sendkey ret
2023/12/29 18:52:56 socat[6854] E connect(5, AF=2 127.0.0.1:4440, 16): Connection refused
```

#### Linux Distribution & Kernel

Run the following and paste the output.

```bash
lsb_release --all
uname -a
```
lsb_release is not installed however os-release is present:

```
didier[bigfiles]$ cat /etc/os-release
NAME=Slint
VERSION="15.0"
ID=slint
VERSION_ID=15.0
PRETTY_NAME="Slint 15.0"
ANSI_COLOR="0;34"
CPE_NAME="cpe:/o:slint:slint_linux:15.0"
HOME_URL="https://slint.fr/"
SUPPORT_URL="https://www.freelists.org/list/slint"
BUG_REPORT_URL="https://www.freelists.org/list/slint"
```
```
didier[bigfiles]$ uname -a
Linux ici.ici.net 6.6.7 #1 SMP PREEMPT_DYNAMIC Thu Dec 14 21:30:24 CET 2023 x86_64 Intel(R) Core(TM) i7-4770 CPU @ 3.40GHz GenuineIntel GNU/Linux
```
FYI this script instead works: https://slint.fr/doc/windows11.sh

lj3954:
Fixes #888

### DIFF
--- a/quickemu
+++ b/quickemu
@@ -972,9 +972,12 @@ function vm_boot() {
         -device usb-redir,chardev=usbredirchardev3,id=usbredirdev3
         -device pci-ohci,id=smartpass
         -device usb-ccid
-        -chardev spicevmc,id=ccid,name=smartcard
-        -device ccid-card-passthru,chardev=ccid
     )
+
+    if "${QEMU}" -chardev spicevmc,id=ccid,name= 2>&1 | grep -q smartcard; then
+        args+=(-chardev spicevmc,id=ccid,name=smartcard
+               -device ccid-card-passthru,chardev=ccid)
+    fi
 
     # setup usb-controller
     [ -z "${USB_CONTROLLER}" ] && USB_CONTROLLER="$usb_controller"
--- a/quickemu
+++ b/quickemu
@@ -977,6 +977,8 @@ function vm_boot() {
     if "${QEMU}" -chardev spicevmc,id=ccid,name= 2>&1 | grep -q smartcard; then
         args+=(-chardev spicevmc,id=ccid,name=smartcard
                -device ccid-card-passthru,chardev=ccid)
+    else
+        echo "WARNING! ${QEMU} was not compiled with support for smartcard devices"
     fi
 
     # setup usb-controller


TuxVinyards:
Hi Liam.  I have been looking at this and #888 this morning ...

Couple of things:

First is the quickemu variable `"$QEMU"`  should be used otherwise we break the ability to switch system type.  ie we should test with `if "$QEMU"-chardev spicevmc,id=ccid,name=`  etc    I can think of instances where this might happen.

Secondly #888 refers to using QEMU 8.2.   I think the real problem lies here.   I don't think 8.2 is really release ready.  I am not sure why Didier is using it?  

New releases of QEMU are well known for being buggy.  `https://qemu.readthedocs.io`  for example routes itself to version 8.2 and ends up with a 404 error.  And `https://readthedocs.org/projects/qemu/builds/`  shows that 8.2 is consistently failing to build.

When I test `qemu-system-x86_64 -chardev spicevmc,id=ccid,name=` against  QEMU 8.0.4 and QEMU 8.1.3 they both show smartcard as available.   So 8.2 may in fact  have the smartcard option, it's just that there's a bug in that release .....

lj3954:
I've fixed it to use the QEMU variable. I disagree that QEMU 8.2 should be a problem. It was released 10 days ago, and is already in the stable repositories of distros such as Arch. The smartcard option is definitely not unsupported in QEMU 8.2. It's just not enabled in that build. You can clearly find in the build log they posted (https://slackware.uk/slint/x86_64/slint-15.0/source/qemu/LOG_8.2.0) that smartcard support is explicitly disabled. I've been using QEMU 8.2 for several days already since one of my partitions has Fedora Rawhide installed on it, and it hasn't given me any issues.

TuxVinyards:
Thanks for the var fix. 

Haven't tried 8.2  yet ...

👍

philclifford (approved):
LGTM

TuxVinyards:
This isn't affecting all distros. My feeling is that issue is down to differing Qemu builds and/or differing dependency builds in different distros.   There was similar problem back in June too, last year  #717 

 Manjaro offered  updates to Qemu 8.2 yesterday.  It all works fine.  Smartcard is present as normal.  

8 2 smartcard

Irrespectively,   I have added a hot fix for this problem to qqX this morning while we wait for a new quickemu update ...

<https://github.com/TuxVinyards/qqX/releases/tag/1.4.05>

lj3954:
That is correct. As I mentioned in my previous comment, the smartcard feature was explicitly disabled in that build of QEMU; the release had nothing to do with it.

TuxVinyards:
It's a neat test that you are using. I'm using exactly the same one in qqX.

The qemu request is passed is grepped for the term 'smartcard'. No?

So, looking closely at the screenshot that I posted Manjaro now qemu 8.2 and passes your test as okay.

Have we got our wires crossed somewhere ??

lj3954:
QEMU 8.2 works just fine as long as it's built with smartcard support, it should pass the test and thus add the arguments. It doesn't matter what the QEMU version is. This check specifically fixes quickemu when QEMU is built without smartcard support. The major distros package it with smartcard support enabled, but some may not and some people build QEMU themselves.

TuxVinyards:
I did give your PR a :+1:  earlier.   It was a good addition, after you put in the variable.

Looks like my original doubts over 8.2 itself were unfounded.   But to say "QEMU 8.2 did not change anything"  would not be right either.  This upstream change has caused things to happen.  And this is what I meant ....

Hopefully both our fixes, yours and my interpretation of yours, will improve the situation.   

As this isn't the first occasion distro maintainers have built Qemu without smartcard enabled ie #717  let me give your PR two extra +1's :+1:  :+1:

TuxVinyards:
I am sure that we would all like to move on from this discussion,  and probably none more so than me.  However, I have just realised that a **_warning is missing_**.  

We do need to let people know what has happened  rather than have them leaving all bewildered and starting to file issues.

The usual quickemu method is to echo something onto the screen:

Screenshot at 2024-01-06 11-07-05

Otherwise still +1 +1 :rocket:

quickemu
```diff
@@ -972,10 +972,13 @@ function vm_boot() {
...
         -device pci-ohci,id=smartpass
         -device usb-ccid
-        -chardev spicevmc,id=ccid,name=smartcard
-        -device ccid-card-passthru,chardev=ccid
     )
 
+    if "${QEMU}" -chardev spicevmc,id=ccid,name= 2>&1 | grep -q smartcard; then
+        args+=(-chardev spicevmc,id=ccid,name=smartcard
+               -device ccid-card-passthru,chardev=ccid)
+    fi
```
philclifford, line 980:
```suggestion
    else
        echo "WARNING! ${QEMU} was not compiled with support for smartcard devices"
    fi
```

philclifford (approved):

DidierSpaier:
I have rebuilt and installed Qemu  8.2.0 with smartcard support, and still I get this error:
```
qemu-system-x86_64: -chardev spicevmc,id=ccid,name=smartcard: unsupported type name: smartcard
allowed spice char type names: vdagent, usbredir
```
[LOG_8.2.0_2.txt](https://github.com/quickemu-project/quickemu/files/14170796/LOG_8.2.0_2.txt)

I attach the full build log.

DidierSpaier:
I applied the firdt patch, then typed:
```

flexiondotorg:
Closed via #986

PR #891, opened 2023-12-29
quickget: show info about operating system WIP

zen0bit:
Since is a lot of supported operating system.
It could be useful show short info about them..

- [x] os_about function
- [x] all supported OS - tried to include only first sentence from homepage, but sometimes I maked my own from about pages etc.
Feel free to EDIT.
- [ ] guickget argument
- [ ] show info somewhere

### DIFF
--- a/quickget
+++ b/quickget
@@ -372,6 +372,95 @@ function os_homepages(){
     echo "${HOMEPAGE}"
 }
 
+function os_about(){
+    local SIMPLE_NAME=""
+    local ABOUT=""
+    SIMPLE_NAME="${1}"
+    case ${SIMPLE_NAME} in
+        alma)              ABOUT="Community owned and governed, forever-free enterprise Linux distribution, focused on long-term stability, providing a robust production-grade platform. AlmaLinux OS is binary compatible with RHEL®.";;
+        alpine)            ABOUT="Security-oriented, lightweight Linux distribution based on musl libc and busybox.";;
+        android)           ABOUT="Port Android Open Source Project to x86 platform";;
+        antix)             ABOUT="Fast, lightweight and easy to install systemd-free linux live CD distribution based on Debian Stable for Intel-AMD x86 compatible systems.";;
+        archlinux)         ABOUT="Lightweight and flexible Linux® distribution that tries to Keep It Simple.";;
+        archcraft)         ABOUT="Yet another minimal Linux distribution, based on Arch Linux.";;
+        arcolinux)         ABOUT="Is all about becoming an expert in linux.";;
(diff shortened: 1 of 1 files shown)


TuxVinyards:
More good work on adding OS info :+1: 

Couple of things that I noticed, that you might want to tweak:

Ubuntu-Kylin text needs to be in English ?

Linux Mint  should probably lead with the phrase: Ubuntu based. 

And Manjaro  should probably lead with:  Arch based.  

Lubuntu should probably say that it is lightweight.  Also that it  features the LXDE/LXQT  desktop.

Same with Ubuntu Mate.  Needs to say Mate desktop somewhere.

Void and Ubuntu-Mate  also need initial Caps

:rocket:

TuxVinyards:
PS noticing the info line on Kali-linux  ...   Should we have Parrot OS on quickget?   It's probably number 2 in the pentest OS's.  Balanced approach ...

zen0bit:
> More good work on adding OS info 👍
> 
> Couple of things that I noticed, that you might want to tweak:
> 
> Ubuntu-Kylin text needs to be in English ? - ? I changed Ubuntu-Kylin and Deepin to eng version since quickemu is only in english (yet - thinking about translations for a while 🤔)
> - ? 
> Linux Mint should probably lead with the phrase: Ubuntu based.
>  - 👍
> And Manjaro should probably lead with: Arch based.
>  - 👍
> Lubuntu should probably say that it is lightweight. Also that it features the LXDE/LXQT desktop.
>  - 👍
> Same with Ubuntu Mate. Needs to say Mate desktop somewhere.
>  - 👍
> Void and Ubuntu-Mate also need initial Caps
>  - 👍

ParrotOS 🤔

TuxVinyards:
Translations would be good but can get very complex.  Like .po files.

Technically you could have a relational array in an auxiliary text file but this again gets very restrictive as things start to become immutable. Something for mature projects only, I think

zen0bit:
> PS noticing the info line on Kali-linux ... Should we have Parrot OS on quickget? It's probably number 2 in the pentest OS's. Balanced approach ...

#893

zen0bit:
Included in #892

philclifford:
This feature is unnecessary and adds a maintenance burden and an additional barrier to contributors.

TuxVinyards:
@philclifford  It's only a one off, one liner?

I think this would be quite good for people browsing the download lists as they wouldn't have to keep opening the OS home pages just to get a basic synopsis.

I will be adding this to qqX irrespectively.

Link maintenance is a lot more work.   😂

All the best 2024 🥂

PR #901, opened 2024-01-08
Improve the code used to determine host RAM amount.

### LINKED ISSUE #899: quickemu create VM error when use 1T RAM
image

image

I think it looks better use `free -g` directly but not use 'free -h' then grep it

temp fix: /usr/bin/quickemu
```
  local RAM_VM="4G"
    # if [ -z "${ram}" ]; then
    #     local RAM_HOST=""
    #     RAM_HOST=$(free --mega -h | grep Mem | cut -d':' -f2 | cut -d'G' -f1 | sed 's/ //g')
    #     #Round up - https://github.com/wimpysworld/quickemu/issues/11
    #     RAM_HOST=$(printf '%.*f\n' 0 "${RAM_HOST}")
    #     if [ "${RAM_HOST}" -ge 128 ]; then
    #         RAM_VM="32G"
    #     elif [ "${RAM_HOST}" -ge 64 ]; then
    #         RAM_VM="16G"
    #     elif [ "${RAM_HOST}" -ge 16 ]; then
    #         RAM_VM="8G"
    #     elif [ "${RAM_HOST}" -ge 8 ]; then
    #         RAM_VM="4G"
    #     fi
    # else
    #     RAM_VM="${ram}"
    # fi
    echo ", ${RAM_VM} RAM"
```

lj3954:
Fixes #899 

Current RAM check is not very versatile. It functions by printing out text following the colon, preceding the first instance of the character 'G', excluding spaces. This will throw an error on systems with under 1GB or over 1000GB of memory, since 'G' does not follow the first value, and it's not possible to round (or use the greater than/equal to operator) on strings. This resolves that issue, as well as removing the need for rounding in the first place, since free --giga does not print out floating point values.

### DIFF
--- a/quickemu
+++ b/quickemu
@@ -313,9 +313,8 @@ function vm_boot() {
     local RAM_VM="2G"
     if [ -z "${ram}" ]; then
         local RAM_HOST=""
-        RAM_HOST=$(free --mega -h | grep Mem | cut -d':' -f2 | cut -d'G' -f1 | sed 's/ //g')
-        #Round up - https://github.com/wimpysworld/quickemu/issues/11
-        RAM_HOST=$(printf '%.*f\n' 0 "${RAM_HOST}")
+        # Determine the number of gigabytes of RAM in the host by extracting the first numerical value from the output.
+        RAM_HOST=$(free --giga | tr ' ' '\n' | grep -m 1 [0-9])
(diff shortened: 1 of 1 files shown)


philclifford (approved):
LGTM

flexiondotorg:
Closed via #986

PR #917, opened 2024-02-04
feat: basic arm64 support for quickget

lj3954:
This should not be merged until quickemu has support for other architectures 

- Check architecture of host system
- Added Ubuntu Desktop 22.04/daily-live arm64
- Added Ubuntu Server arm64 
- Added Fedora arm64 (Certain variants only, falls back on x86_64 otherwise)
- Append ISO architecture to config file (excluding amd64) for use in quickemu

### DIFF
--- a/quickget
+++ b/quickget
@@ -372,6 +372,16 @@ function os_homepages(){
     echo "${HOMEPAGE}"
 }
 
+function check_arch() {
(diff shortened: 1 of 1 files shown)


lj3954:
This PR is no longer relevant. I'm putting effort towards gathering images of aarch64 (& riscv64) media in https://github.com/quickemu-project/quickget_configs, and replacing quickget with a tool which uses said configurations (#1321). #1290 may implement some of what was proposed here.

PR #929, opened 2024-02-15
feat: OS distribution rss feed

zen0bit:
For now DistroWatch only for testing. Could be any number of RSS feeds

Should I work more on this?
usecase: Get actual new release info into github action, test and apply to quickget without intervention.

see [issues](https://github.com/oSoWoSo/quickemu/issues)
and
[actions](https://github.com/oSoWoSo/quickemu/actions/workflows/rss.yml)

### DIFF
--- a/.github/workflows/rss.yml
+++ b/.github/workflows/rss.yml
@@ -1,0 +1,27 @@
+--- # RSS feed
+name: Monitor new OS versions
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


flexiondotorg:
I don't think merging this is wise, it will spam the issue tracker. I'd much prefer to curate distros and add support for the ones in demand and have people interested in maintaining.

PR #940, opened 2024-02-21
quickemu: Support $disk_format var for $disk_img

gmt4:
This adds disk_format=${disk_format:-qcow2} variable that defaults to qcow2 format, so the vm.conf files can use $disk_format to specify the format of the $disk_img when this is different from the default qcow2.

### DIFF
--- a/quickemu
+++ b/quickemu
@@ -1145,7 +1145,7 @@ function vm_boot() {
     else
         # shellcheck disable=SC2054,SC2206
         args+=(-device virtio-blk-pci,drive=SystemDisk
-            -drive id=SystemDisk,if=none,format=qcow2,file="${disk_img}" ${STATUS_QUO})
+            -drive id=SystemDisk,if=none,format=${disk_format:-qcow2},file="${disk_img}" ${STATUS_QUO})
     fi
 
     # https://wiki.qemu.org/Documentation/9psetup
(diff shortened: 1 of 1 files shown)


lj3954:
Could you not use the file extension in the disk_img variable to determine the correct format?

gmt4:
> Could you not use the file extension in the disk_img variable to determine the correct format?

yep, thanks for the suggestion. Instead of guessing, I just went with asking for qemu-img info to do the guessing for me.

TuxVinyards:
> ` | gawk '/^file format: / {print $NF}' `

` | grep 'file format' | cut -d ':' -f 2 | tr -d ' ' `    

 Not everyone has gawk installed.   I would suggest this to be more portable.  It's also easier to understand ...

gmt4:
Done!, thanks for spotting it, i just went with gawk as I saw another use of gawk in https://github.com/quickemu-project/quickemu/blob/master/quickemu#L1411

TuxVinyards:
> @TuxVinyards  Done!, thanks for spotting it, i just went with gawk as I saw another use of gawk in https://github.com/quickemu-project/quickemu/blob/master/quickemu#L1411

Two wrongs don't make a right.  Line 1411 has bad use of 'cat' as well. 😵‍💫

flexiondotorg:
Closed via #986

Issue #944, opened 2024-02-23
Kali Linux out of disk space


Parrot7483:
#### Expected behaviour

Following the installation process with default settings and ensuring all software upgrades are completed, the system should not encounter disk space limitations.

#### Actual behaviour

Out of disk space after default install.

#### Steps to reproduce the behaviour

Install kali linux current with all default settings and upgrade software afterwards

#### Proposed Change

Increase the default disk size to 20G - 30G

flexiondotorg:
Closed via #986

Issue #738, opened 2023-08-29
Gentoo VMs do not have enough disk space.


jon-bit:
When you make the disk image for it, it is not enough to install Gentoo. I don't know how much space is needed for Gentoo(I tried in installing it last night for the first time) but I keep getting the no more disk space error. Is it meant only to be booted live or is their a real/temp fix for this?

zen0bit:
I can easily solve that.

But leaving for now for someone new..
Since I already solved my first issue here 😃

PR #947, opened 2024-02-25
quickemu default disk size increse

guyosi:
this increases the default disk size from 16 to 32, fixing issues #738  and #944


flexiondotorg:
Closed via #986

PR #949, opened 2024-02-27
Update README.md

dsbrown:
The instructions for git clone is missing one step: sudo make install. This required me to examine the makefile targets and figure it out for myself. Not so bad for me, but others may be confused about what to do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,7 @@ sudo apt install quickemu
 ``` bash
 git clone --filter=blob:none https://github.com/wimpysworld/quickemu
 cd quickemu
+sudo make install
 ```
 
 Now install all the **Requirements** documented above.


lj3954:
The makefile is not in this repository. You running `sudo make install` after cloning this repository will just throw an error. Anyways, one could just run quickemu directly out of whichever directory they cloned it into, either by pointing directly to it or adding it to their PATH.

philclifford:
The Makefile is in the docs subdirectory. It is provided as a courtesy/convenience not as a recommended installation method.
If it is to be used the instruction should be to install all the documented pre-requisites (plus `pandoc` if the man pages are to be regenerated) then

```
cd docs
make
sudo make install
```

flexiondotorg:
Closed via #986

PR #969, opened 2024-03-22
Add spice-gtk-tools to Fedora packages in README.md

fxkr:
On Fedora, the spice-gtk-tools provides the spicy command.

Without it, quickemu with default config fails with the following message:

```
ERROR! Requested 'spicy' as viewer, but 'spicy' is not installed.
```


fxkr:
Wait, it was already there. Huh. Sorry, ignore this pull request.

PR #981, opened 2024-04-05
Fix USB input devices after macOS Sonoma update

chasecovello:
A recent macOS Sonoma update included a change to the handling of USB controllers that made the virtual keyboard and mouse inoperative. Changing the USB controller from ehci to xhci fixes it. This change also works for Big Sur and later, so the USB controller is set to xhci for Big Sur through Sonoma.


flexiondotorg:
Closed via #986

PR #990, opened 2024-04-11
readd EasyOS

zen0bit:
Dont trash my work pls


flexiondotorg:
This does not work. The reason I removed easyOS is because the download site no longer exists.

image

flexiondotorg:
It looks like that domain is in a DNS blacklist and being filtered on my network.

PR #995, opened 2024-04-12
fix blendOS downloads

TuxVinyards:
VM PATH needed a tweak

### DIFF
--- a/quickget
+++ b/quickget
@@ -3434,7 +3434,14 @@ fi
 
 if [ -n "${2}" ]; then
     RELEASE="${2,,}"
-    VM_PATH="${OS}-${RELEASE}"
+
+    if [[ $OS == "blendos" ]]; then
+        BlendRelease=${RELEASE/'/'/'-'}
+        BlendRelease=${BlendRelease/168/}
+        VM_PATH="${OS}-${BlendRelease}"
+    else
+        VM_PATH="${OS}-${RELEASE}"
+    fi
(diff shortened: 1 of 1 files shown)


flexiondotorg (changes requested):
These changes for blendos should be in a function specific to blendos, not in the general code path.

TuxVinyards:
> These changes for blendos should be in a function specific to blendos, not in the general code path.

I did try that.  The trouble is, that  `$RELEASE` contains a forward slash and `web_get`  needs to have things presented to it in that fashion or it won't download.

 The standard  `$VM_PATH`   assembly uses  `$RELEASE`  as  component,  so we have to change things at that point, after the downloads but before writing to hard drive.  If we don't change it, we end up with a broken distro name and sub folders being created, with the .conf file inside the main VM folder.

Screenshot at 2024-04-13 13-52-11

So, in the same way that both Manjaro & Fedora have to have special entries in that section, we unfortunately have to do that here too .....

lj3954:
I think we can easily use this data to separate it into releases & editions instead. Also, a temp file definitely isn't needed to store this plain text data. I'll have a PR up for these changes in a little bit.

TuxVinyards:
Good work Liam.  I woke up thinking along similar lines this morning.  Good editions filter.

Hopefully no grumbles from @lj3954  on this being a combined effort   :rofl:

PR #1004, opened 2024-04-14
Use only latest blendos release

lj3954:
Currently, quickget determines the blendos release/edition combinations using its RSS feed, and then prints them all out as releases. For example, you'll see output like this: `plasma/1688424625 gnome/1688424211 deepin/1688423865 lxqt/1688423615 mate/1688423313 xfce/1688423053 cinnamon/1688422684 plasma/168787346`.
This also causes significant issues, such as directory names including slashes, as seen in #995.

This PR proposes replacing this mess with just offering the user the latest release, and listing out the available editions instead. The latest release is determined by numerically sorting through every available release within the selected edition. The releases & editions cannot be listed separately, since each edition has different releases (which are based on epoch time). Therefore, I think this is the best available solution.

Closes #995

### DIFF
--- a/quickget
+++ b/quickget
@@ -524,13 +524,15 @@ function releases_biglinux() {
 }
 
 function releases_blendos() {
-    # Pull the rss feed
+    echo latest
+}
+
+function editions_blendos() {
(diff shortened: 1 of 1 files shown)


zen0bit (approved):

TuxVinyards:
@flexiondotorg  This looks all okay to me now.   The  'latest '  filter now works ....  :+1:

PR #1005, opened 2024-04-14
Blendos fix3

TuxVinyards:
blendos-fix3  makes sure that it s actually the latest , not just the one at the top of the rss list

-  Combined efforts  :+1: 

Resolves #995

Improves on solution #1004  by using sort instead of head -n 1

Removes 'latest_blendos_release' as this only returned null.  Good idea but probably wasn't needed.

fix3-works

### DIFF
--- a/quickget
+++ b/quickget
@@ -1565,14 +1565,11 @@ function get_biglinux() {
 function get_blendos() {
     local HASH=""
     local URL=""
-    local latest_blendos_release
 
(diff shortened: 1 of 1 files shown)


flexiondotorg:
I'll close this and request that @lj3954 and @TuxVinyards collaborate on #1004 to propose one pull request that addresses issues with BlendOS.

TuxVinyards:
@lj3954   Github's twists and turns drive me up the wall.   Can you sort this?    Github said this request was merge-able when I put it up.

Maybe @flexiondotorg  is just wanting your seal of approval?

PR #1015, opened 2024-04-15
feat: test_isos function

zen0bit:


TuxVinyards:
```
            elif [[ "${OS}" == *"ubuntu"* ]]; then
                URL=$(get_ubuntu)
```

I think this is where the problem I spotted earlier comes from.  This won't call  get_kubuntu,  lubuntu , xubuntu etc

flexiondotorg (approved):

flexiondotorg (changes requested):
Please see the comment from @TuxVinyards

PR #1024, opened 2024-04-15
ci: quickget tests

zen0bit:
Signed-off-by: zenobit <zenobit@disroot.org>


flexiondotorg (changes requested):
This will not function until the new TUI is fixed:

- https://github.com/quickemu-project/quickemu/pull/999#issuecomment-2058550075

Issue #1049, opened 2024-04-16
bug: Quickget response to partial OS request where OS has Editions is broken


philclifford:
Expected : 

``` text
$ ./quickget -s debian 
ERROR! You must specify a release.
 - Releases: 10.0.0 10.1.0 10.2.0 10.3.0 10.4.0 10.5.0 10.6.0 10.7.0 10.8.0 10.9.0 10.10.0 10.11.0 10.12.0 10.13.0 11.0.0 11.1.0 11.2.0 11.3.0 11.4.0 11.5.0 11.6.0 11.7.0 11.8.0 11.9.0 12.0.0 12.1.0 12.2.0 12.3.0-NEVER-RELEASED 12.4.0 12.5.0
 - Editions: standard cinnamon gnome kde lxde lxqt mate xfce netinst

```

``` text
$ ./quickget -s debian 
ERROR! You must specify a release.
 - Editions: standard cinnamon gnome kde lxde lxqt mate xfce netinst

$ ./quickget -s debian mate
ERROR! Debian mate is not a supported release.

10.0.0 10.1.0 10.2.0 10.3.0 10.4.0 10.5.0 10.6.0 10.7.0 10.8.0 10.9.0 10.10.0 10.11.0 10.12.0 10.13.0 11.0.0 11.1.0 11.2.0 11.3.0 11.4.0 11.5.0 11.6.0 11.7.0 11.8.0 11.9.0 12.0.0 12.1.0 12.2.0 12.3.0-NEVER-RELEASED 12.4.0 12.5.0$
```

Note we were left with a shell prompt at the end of a long line there ---------->----------->----------->----------->----------->-----------> ... 

and if we provide the right inputs in the right order

``` text

$ ./quickget -s debian 12.5.0 mate
https://cdimage.debian.org/debian-cd/12.5.0-live/amd64/iso-hybrid/debian-live-12.5.0-amd64-mate.iso

```

Same for `alba`, `linuxmint`, `rockylinux` `zorin`  etc. but  not `ubuntu*` `windows` ot `macos` so presumably it is always and only where there are EDITIONS present.

PR #1054, opened 2024-04-16
Fix regression in the case of a missing release

lj3954:
The os_error_release function only prints the releases if an editions function does not exist. This fixes the error.


lj3954:
Fixes #1049

Issue #1071, opened 2024-04-17
I am trying to install Monterey on Ubuntu


petrasvestartas:
#### Expected behaviour
Works

#### Actual behaviour

I installed quickmenu and gui and monterey like this:

```bash
sudo apt-add-repository ppa:flexiondotorg/quickemu
sudo apt update
sudo apt install quickemu

sudo add-apt-repository ppa:yannick-mauray/quickgui
sudo apt update
sudo apt install quickgui

quickget macos monterey
quickemu --vm macos-monterey.conf

```

https://github.com/quickemu-project/quickemu/assets/18013985/634a2ec5-0675-4ec1-9fd9-6de45db84343

![Screenshot_20240417_163623_Gallery](https://github.com/quickemu-project/quickemu/assets/18013985/5a8303bf-3890-41e9-ae89-cd4c54b748a1)

TuxVinyards:
@flexiondotorg we so need a wiki for new users

@petrasvestartas please see the qqX FAQs > My VM won't work ....

lj3954:
macOS Monterey is known to have many issues in quickemu. Catalina is the recommended release if the software you need is compatible with that release, and Ventura should be relatively stable as well if you need something newer.

petrasvestartas:
@lj3954 Even for ventura, could are my steps for installation are correct?

TuxVinyards:
Here https://github.com/TuxVinyards/qqX/wiki/FAQs-and-Help#macos-hangs-when-running-the-setup  

And the link at the bottom ....

petrasvestartas:
@TuxVinyards When he says: Using Quickemu from the master (what will become 4.9.3) I have successfully installed Ventura.

What does it actually mean?

PR #1073, opened 2024-04-18
Add help notes for New Users

TuxVinyards:
Attempt to de-clog quickemu issues ...

See #1071 as recent example


flexiondotorg:
@TuxVinyards Thanks for this, but I'm going to close it. Looking at the README.md, it is trying to do too much. I think the best approach is to start to move sections of the READNE to wiki pages.

TuxVinyards:
@flexiondotorg  Yes.  Agree. The Quickemu readme has become an extremely long single page. 

I look forward to seeing the results.

PR #1087, opened 2024-04-20
Make curl based web handlers and update all ISO downloaders

flexiondotorg:
This pull request adds some new web handlers that use `curl` by default:

- `web_pipe()`: Download a file from the web and pipe it to stdout
- `web_check()`: Checks if a URL is reachable; an alternative to `wget --spider`
- `web_redirect()`: Checks if a URL needs to be redirected and returns the final URL
- `web_get()`: Already existed to download a file from the web, but now uses `curl` by default

## Notes

- `web_get()` will fallback to `wget2` and `wget` if `curl` fails. 
- Most ad-hoc calls to `wget` have been replaced with calls to the appropriate `web_*()` handlers.
  - A separate pull request will be submitted that replaces `wget --spider` with `web_check()`.
- Downloads have been manually tested and releases/editions updated where appropriate.
- Several `releases_*()` and `editions_*()` scripts have been made dynamic.
- URL redirection is consistently handled now, making downloads much more reliable; particularly from Sourceforge, dynamic views of S3 buckets (NixOS), GitHub and anything using [Content-Disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition).

## Rationale

- `aria2c` behaves inconsistently across distributions depending on how it was compiled.
- Some distributions are starting to symlink `wget2` to `wget`, but they do not behave the same way.
- `curl` has stable interfaces, so create functions in `quickget` to establish common ways to fetch/download content from the web.
- `curl` can present a consistent download progress bar.
- `curl` has been used throughout `quickget` in an ad-hoc manner, but was not a documented requirement and is not installed by default in the `quickget` packages for a number of distros.
- Creating some common handlers as the only way to access web content provides a smaller debug and maintenance surface.

### DIFF
--- a/quickget
+++ b/quickget
@@ -1448,8 +1448,7 @@ function web_pipe() {
 
 # Download a file from the web
 function web_get() {
-    local URL=""
-    URL=$(web_redirect "${1}")
+    local URL="${1}"
     local DIR="${2}"
     local FILE=""
     local USER_AGENT="Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/123.0.0.0 Safari/537.36"
@@ -1459,6 +1458,9 @@ function web_get() {
     else
         FILE="${URL##*/}"
     fi
+
+    # Process any URL redirections after the file name has been extracted
+    URL=$(web_redirect "${URL}")
 
     while (( "$#" )); do
         if [[ $1 == --header ]]; then
--- a/quickget
+++ b/quickget
@@ -809,7 +809,7 @@ function editions_artixlinux() {
 }
 
 function releases_athenaos() {
-    web_pipe "https://api.github.com/repos/Athena-OS/athena/releases" | grep 'download_url' | grep rolling | cut -d'/' -f8 | uniq | tr '\n' ' '
+    echo $(web_pipe "https://api.github.com/repos/Athena-OS/athena/releases" | grep 'download_url' | grep rolling | cut -d'/' -f8 | uniq | tr '\n' ' ')
 }
 
 function releases_batocera() {
--- a/quickget
+++ b/quickget
@@ -946,7 +946,7 @@ function releases_elementary() {
 }
 
 function releases_endeavouros() {
-    local ENDEAVOUR_RELEASES="$(curl -s https://mirror.alpix.eu/endeavouros/iso/ | LC_ALL="en_US.UTF-8" sort -Mr | grep -o -P '(?<=<a href=").*(?=.iso">)' | grep -v 'x86_64' | cut -c 13- | tr '\n' ' ')"
+    local ENDEAVOUR_RELEASES="$(web_pipe "https://mirror.alpix.eu/endeavouros/iso/" | LC_ALL="en_US.UTF-8" sort -Mr | grep -o -P '(?<=<a href=").*(?=.iso">)' | grep -v 'x86_64' | cut -c 13- | head -n 5 | tr '\n' ' ')"
     echo "${ENDEAVOUR_RELEASES,,}"
 }
 
--- a/quickget
+++ b/quickget
@@ -1147,11 +1147,11 @@ function releases_oraclelinux() {
 }
 
 function releases_parrotsec() {
-    echo 6.0 5.3 4.11.3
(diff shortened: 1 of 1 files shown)


philclifford:
I'm guessing the extended dynamism, which is **excellent** from our pov in terms of robustness and maintenance burden, is the main reason `list_csv` has gone from 17.4 seconds (potato laptop and mobile 4G hotspot ...) to 42 seconds with only 22 extra/fixed results.

I can put the kettle on while `cog` builds the docs but `quickgui` might suffer more on the usability front and may need some tlc.    


``` shell
#HEAD of master 
time ./quickget list |wc -l
1028

real	0m17.466s
user	0m2.849s
sys	0m0.562s

time ./quickget list_csv |wc -l
1028

real	0m17.418s
user	0m2.787s
sys	0m0.522s


time ./quickget list_json |wc -l
9245

real	0m21.328s
user	0m2.810s
sys	0m0.540s

```

``` shell
# web_pipe branch

time ./quickget list_csv
real	0m40.349s
user	0m7.206s
sys	0m0.889s

time ./quickget list_csv |wc -l
1050

real	0m41.975s
user	0m7.343s
sys	0m0.893s

time ./quickget list |wc -l
1050

real	0m58.758s
user	0m7.162s
sys	0m0.941s

time ./quickget list_json |wc -l
9443

real	0m45.488s
user	0m7.504s
sys	0m1.000s

```

philclifford:
Helpful iso info is no longer functional although still being offered:

(This is broken on HEAD of main btw, not by this PR so I'll push this comment to a new issue)

```
# master and web_pipe
 ./quickget 
ERROR! You must specify an operating system.
 - Operating Systems: alma [...]zorin

        You can also use these arguments:
        Only show ISO download URL
            --show-iso-url / -s {distro} {release} [edition]
        Test if ISO is available
            --test-iso-url / -t {distro} {release} [edition]
        Open distro homepage
            --open-distro-homepage / -o {distro}
        Only download ISO
            --download-iso / -d {distro} {release} [edition]

# 4.9.1 -ish 
~/src/quickemu$ quickget --show-iso-url debian 12.5.0
https://cdimage.debian.org/debian-cd/12.5.0-live/amd64/iso-hybrid/debian-live-12.5.0-amd64-standard.iso

#HEAD of master and web_pipe
~/src/quickemu$ ./quickget --show-iso-url debian 12.5.0
ERROR! --show-iso-url is not a supported OS.

alma [...] zorin

```

same for long and short -s -t -d options 

--help -h doesn't work (yet ?) either, but neither do most of the parameters it would list so ...

flexiondotorg:
I suspect the slower render time is due to the robust URL redirection handling and does not have so much to do with the increase in dynamic releases/editions. Every URL lookup happens at least twice now.

It is possible I can optimise this, but for now, I think I'll go with the "optimise last" pattern and make sure everything works reliably first. I have more dynamic additions and improvements in the works too.

flexiondotorg:
@philclifford For some local per testing you can try editing the `web_pipe()` function in `quickget` to just be this:

```shell
function web_pipe() {
    curl --silent --location "${1}"
}
```

That :point_up: should still follow the redirection but not be as slow.

zen0bit:
> --help -h doesn't work (yet ?) either, but neither do most of the parameters it would list so ...

run script not yet modified...

zen0bit:
Should I make my changes against changes from web_pipe branch then?
Or still from master? Don't wanna rebase again...

lj3954:
> I suspect the slower render time is due to the robust URL redirection handling and does not have so much to do with the increase in dynamic releases/editions. Every URL lookup happens at least twice now.
> 
> It is possible I can optimise this, but for now, I think I'll go with the "optimise last" pattern and make sure everything works reliably first. I have more dynamic additions and improvements in the works too.

It absolutely also does have to do with the dynamic release/edition detection. Requests have to be made for every distro sequentially, which of course leads to a long wait. I think the most important thing that needs to be done to improve the performance of list_csv is to parallelize it. I've proposed some changes in #1089, take a look and see if that works.

flexiondotorg:
> Should I make my changes against changes from web_pipe branch then?

Yes :+1: That said, I am about to merge this branch as I think it is looking pretty solid now.

PR #1093, opened 2024-04-21
feat: Fill empty fields in info with ' - '

zen0bit:


TuxVinyards:
I think it will be worth it when it's all done.  :+1: 

We have come a long way from discussions on ways forward :rocket: 

There's a small alphabetical on PureOS and PrimTux I noticed

zen0bit:
why you already not make PR then if you spot it?

What about help us bit more instead of just commenting?

TuxVinyards:
> why you already not make PR then if you spot it?
> 
> What about help us bit more instead of just commenting?

Sorry if it comes over that way.  I have been doing PR's  but you are moving so fast at the moment and there are so many branches all over the place, it gets a bit complicated.  At least for my poor brain :rofl:

GitHub gets me confused at the best of times.

PR #1095, opened 2024-04-21
chore: Move PureOS in os_info

zen0bit:


philclifford:
Sorry, I notice your re-order correction after burying another in my PR.   Mine also changes the description a little as it is closer to theirs I think in placing the target primarily on the children and only secondarily on the hardware.

zen0bit:
Don't be sorry...

PR #1098, opened 2024-04-21
docs: begin linking to wiki

philclifford:
some alignments with wiki contents made
and also some corrections for newer `quickget` options and use of `cog` 

Also includes `curl` dependency in the manual installation examples. 

Will aim to track further commits and wiki development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,9 +91,9 @@ virtual machines.
 -   EFI (with or without SecureBoot) and Legacy BIOS boot
 -   Graphical user interfaces available
 
-## Installation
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -342,17 +372,17 @@ fully functional for all operating systems, including Windows and macOS.
 -   `archlinux` (Arch Linux)
 -   `arcolinux` (Arco Linux)
 -   `artixlinux` (Artix Linux)
--   `athenaos` (Athenaos)
+-   `athenaos` (Athena OS)
```
philclifford, line 375: These changed arise from `cog` : presumably the revised approach to pretty_name needs some additional data or a tweak

PR #1106, opened 2024-04-23
docs: trim some information that is now in the wiki

philclifford:
First effort at removing some weight from the docs and man pages that is now in the linked wiki.


philclifford:
I'll leave this draft for now in case you want to suggest more serious knife-wielding and also so that this evening another regen could also pick up any further changes.

philclifford:
``` diff
-   `tuxedo-os` (Tuxedo OS)
-   `tuxedoos` ()
```

tuxedoos not found in os_info() so PrettyName missing

PR #1109, opened 2024-04-23
Change tuxedo-os to tuxedo

TuxVinyards:
doos are things that doggies do
and that people stand in :rofl:

either tuxedo-os  or tuxedo

try tuxedo, we don't have fedoraos or debianos

but then we do have ubuntu-mate ??

### DIFF
--- a/README.md
+++ b/README.md
@@ -440,7 +440,7 @@ Further information is available from the project
 -   `trisquel` (Trisquel)
 -   `truenas-core` (TrueNAS Core)
 -   `truenas-scale` (TrueNAS Scale)
--   `tuxedo-os` (Tuxedo OS)
+-   `tuxedo` (Tuxedo OS)
 -   `vanillaos` (Vanilla OS)
 -   `void` (Void Linux)
 -   `vxlinux` (VX Linux)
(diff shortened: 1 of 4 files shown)


zen0bit:
I like just = tuxedo
But they use name = tuxedo os
tuxedo-os = looks strange to me
tuxedoos = That poop ting
But we have also vanillaos centos kalibrios and lot of others ...os

### Leaving choice on @flexiondotorg ...

But if you already bite to ubuntu variants
They really should be named like that?
ubuntu
ubuntu-budgie
ubuntu-mate
ubuntu-server
ubuntu-unity
ubuntucinnamon
ubuntukylin
ubuntustudio

Better (for my eyes) will be:
ubuntu
ubuntu-budgie
ubuntu-cinnamon
ubuntu-kylin
ubuntu-mate
ubuntu-server
ubuntu-studio
ubuntu-unity

Consistency everywhere please
That are all the same ubuntu variant just with different desktop no?
 (With exception ubuntu-studio)
Let's call them with same style no?

zen0bit:
I already suggest leave linux from OS names, but not accepted, so I don't even start with leaving os from end of names.. 🤐

PS: Who ever said I use ArchLinux BtW?
    I heard every time just Arch...
🤔

flexiondotorg:
Thanks for the suggestion, but I won't merge this one as the project is Tuxedo OS.

PR #1110, opened 2024-04-23
feat: Add nwg-shell

zen0bit:
Live ISO based on Arch.
To test out nwg-shell on hyprland or sway


quickget
```diff
@@ -919,6 +920,11 @@ function editions_nixos(){
     echo minimal plasma5 gnome
 }
 
+function releases_nwg-shell() {
+    local NWG_RELEASES=$(web_pipe "https://sourceforge.net/projects/nwg-iso/rss?path=/" | grep 'url=' | grep '64.iso' | cut -d'/' -f12 | cut -d'-' -f3 | tr '\n' ' ')
```
flexiondotorg, line 924: Please change this to one line:

```
echo $(web_pipe "https://sourceforge.net/projects/nwg-iso/rss?path=/" | grep 'url=' | grep '64.iso' | cut -d'/' -f12 | cut -d'-' -f3)
```

flexiondotorg (changes requested):

flexiondotorg (approved):

PR #1107, opened 2024-04-23
Finall Pack + CI + solved shellcheck errors

zen0bit:
# New UI finalization

1. Added quickget tests to CI
2. Test made parallel
3. Added SHELL check
4. Added YAML check
5. Fixed introduced regression
6. Finalized arguments and help message
7. Solved shellcheck errors (SC2317 disabled globally)

tests:
1. List all supported OS with releases and editions
8. List infos about all supported OS
9. List all URLs of supported ISOs
10. Test all supported ISOs availability
[Test quickget 🧪](https://github.com/oSoWoSo/quickemu/actions/workflows/test-quickget.yml)
Tests also save results to artifacts for later use (txt in zip)
see it in checks
PS: With parallelization we are under 10 minutes (from 30+)

# Failed:
FAIL: arcolinux-v24.02.01-large           https://mirror.accum.se/mirror/arcolinux.info/iso/v24.02.01/arcolinuxl-v24.02.01-x86_64.iso
FAIL: arcolinux-v24.02.01-small           https://mirror.accum.se/mirror/arcolinux.info/iso/v24.02.01/arcolinuxs-v24.02.01-x86_64.iso
FAIL: arcolinux-v24.05.11-large           https://mirror.accum.se/mirror/arcolinux.info/iso/v24.05.11/arcolinuxl-v24.05.11-x86_64.iso
FAIL: arcolinux-v24.05.11-small           https://mirror.accum.se/mirror/arcolinux.info/iso/v24.05.11/arcolinuxs-v24.05.11-x86_64.iso
FAIL: blendos-v3-gnome                    https://mirror.ico277.xyz/blendos/gnome/blendos-v3-stable-gnome.iso
FAIL: fedora-39-Onyx
FAIL: fedora-39-Sericea
FAIL: rebornos-latest

## Closes #1104 #1150 #1151 #1158

# Help message:
help message

PR #1112, opened 2024-04-25
Implement simple custom_config

lj3954:
Only image types img & iso are supported by quickemu, so throw an error if the file/url doesn't end in one of those. Quickget shouldn't require the user to specify the OS, so instead attempt to determine it from the file (vast majority will be Linux-based anyways). make_vm_config had to be slightly modified to allow for the possibility of full paths to image files.

### DIFF
--- a/quickget
+++ b/quickget
@@ -353,20 +353,48 @@ function list_csv() {
 }
 
 function create_config() {
-    local OS="${1}"
-    local DESTINATION="${2}"
-    local USE
+    local VM_PATH="${1}"
+    local INPUT="${2}"
+    OS="custom"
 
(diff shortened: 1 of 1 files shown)


zen0bit:
I like it but, just keep same structure also for custom VM

Only question is in "local mode".
If image should be moved or copied from previous location

I am for moving (saving space)

zen0bit:
Bit modified your create_config function in my PR #1107 
Check if I messed something...

flexiondotorg (changes requested):
Please can you rebase against current master

flexiondotorg:
CLosed via https://github.com/quickemu-project/quickemu/pull/1143

Issue #1108, opened 2024-04-23
bug: web_check is slow and freezes when no connection


TuxVinyards:
`--check          (-c) <os> [re] [ed] : Check download an OS release/edition is available`

@flexiondotorg  This new API function needs a look at.  It actually goes round the loop no less that three times.

I have spent a couple of hours trying quick fixes eg `--retry 1 --connect-timeout .5 ` etc  but without success.

There's a long delay before even the "Test availability"  message shows.   And even if it the message is placed to precede the first call to `web_redirect`., it still happens.  I haven't yet found where that delay comes from.

Following this, `--check`  goes on to trigger  ` if  [ "${just}" == 'test ' ] `  which then calls  `web_check`  which in turn calls `web_redirect` for the second time ...

At the end of these two checks, `web_check` then does it's own personal third check.

And if  the internet is down, the whole thing it freezes.

If I spent some time on this, I could do a re-write. But being that you have written a lot of this, I wouldn't want to impinge or to put time into a PR that you wouldn't want.

It's a pity that wget2 is causing issues.  `wget --spider ` has been so easy to use, up till now.  :shrug:

TuxVinyards:
Probably what you discovered  `wget --spider`  doesn't work on Microsoft homepages :rofl:

But I replaced `--fail`  and this works better:

```
if curl --location --silent --head --output /dev/null --retry 1 --connect-timeout 1  "https://www.microsoft.com/en-us/windows/" ; then echo yes; else no; fi 

```

PR #1115, opened 2024-04-25
Refactor code for more consistency between functions

Jai-JAP:
the changes need to be reviewed and tested to avoid breakage

### DIFF
--- a/quickget
+++ b/quickget
@@ -35,9 +35,8 @@ function cut_1() {
 }
 
 os_info() {
-    local SIMPLE_NAME=""
+    local SIMPLE_NAME="${1}"
     local INFO=""
-    SIMPLE_NAME="${1}"
     case ${SIMPLE_NAME} in
       #name)            INFO="PrettyName|BasedOf|Credentials|Homepage|Info";;
       alma)             INFO="Alma Linux|Fedora,RedHat|-|https://almalinux.org/|Community owned and governed, forever-free enterprise Linux distribution, focused on long-term stability, providing a robust production-grade platform. AlmaLinux OS is binary compatible with RHEL®.";;
@@ -277,17 +276,15 @@ handle_missing() {
 }
 
 function validate_release() {
-    local DISPLAY_NAME=""
     local RELEASE_GENERATOR=""
-    local RELEASES=""
 
-    DISPLAY_NAME="$(pretty_name "${OS}")"
+    local DISPLAY_NAME="$(pretty_name "${OS}")"
     case ${OS} in
       *ubuntu-server*) RELEASE_GENERATOR="releases_ubuntu-server";;
       *ubuntu*) RELEASE_GENERATOR="releases_ubuntu";;
       *) RELEASE_GENERATOR="${1}";;
     esac
-    RELEASES=$(${RELEASE_GENERATOR})
+    local RELEASES=$(${RELEASE_GENERATOR})
     error_not_supported_release
 }
 
@@ -298,15 +295,15 @@ function list_json() {
 }
 
 function list_csv() {
-    local DISPLAY_NAME
+    local DISPLAY_NAME=""
     local DL=""
-    local DOWNLOADER
+    local DOWNLOADER=""
     local FUNC
-    local OPTION
+    local OPTION=""
     local OS
-    local PNG
+    local PNG=""
     local RELEASE
-    local SVG
+    local SVG=""
     local HAS_ZSYNC=0
 
     # Check if zsync is available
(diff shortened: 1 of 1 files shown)


TuxVinyards:
This is _a very ambitious task_ and I have to admit to having been very sceptical about some of the changes, for example the unexplained  removal of  `  | tr '\n' ' ' ` from the ends of some of the `web_pipe` lines looked destined to cause problems. 

In reality, my initial _tests show speed improvements_ when I switched between current 'master' and your branch, which is good :+1:  

But _I do have concerns_ that there are just soooo many changes here, all at once, and that _the testing task at at one sitting is far too big._ I think this PR needs breaking up into manageable sections or we could end with errrors kreeeping through ....

There are over a thousand shellcheck errors and warnings  currently showing in our beloved scripts.  Somehow it all muddles through and gets there in the end.  Goodness know how it does it, but it somehow manages to do so, just about. The whole thing of editing it is a minefield :rofl:

Maybe placing changes to the way data gets parsed, like removals of  `  | tr '\n' ' ' `  should be placed into separate PRs?

Something non-system breaking  like the removal of  "useless echoes"  SC2005  could be placed in another.   
Eg ` echo $(web_pipe ` is  just  wasted system call,  given that 'web_pipe' already does the echo for us?

jai-pr

Jai-JAP:
> This is _a very ambitious task_ and I have to admit to having been very sceptical about some of the changes, for example the unexplained  removal of  `  | tr '\n' ' ' ` from the ends of some of the `web_pipe` lines looked destined to cause problems. 
> 

No it's not, because in subshells the newlines are automatically changed to spaces so it literally makes no difference.
 
> But _I do have concerns_ that there are just soooo many changes here, all at once, and that _the testing task at at one sitting is far too big._ I think this PR needs breaking up into manageable sections or we could end with errrors kreeeping through ....
> 
> There are over a thousand shellcheck errors and warnings  currently showing in our beloved scripts.  Somehow it all muddles through and gets there in the end.  Goodness know how it does it, but it somehow manages to do so, just about. The whole thing of editing it is a minefield :rofl:
> 
> Maybe placing changes to the way data gets parsed, like removals of  `  | tr '\n' ' ' `  should be placed into separate PRs?
> 
> Something non-system breaking  like the removal of  "useless echoes"  SC2005  could be placed in another.   
> Eg ` echo $(web_pipe ` is  just  wasted system call,  given that 'web_pipe' already does the echo for us?
> 
It was a bit confusing whether to do the task this way because it would mean many small PRs with literally one or two commits all aiming to improve code quality without any usability changes and instead we can test these changes here in a draft PR. It may be a bit difficult to manage but it is better than creating many smaller PRs. 

Anyway I am fine with doing it anyway.

lj3954:
> In reality, my initial _tests show speed improvements_ when I switched between current 'master' and your branch, which is good 👍

It shouldn't. Exactly the same requests are being made either way. Even if the string manipulation was more efficient in this branch, it would make a difference of a couple milliseconds, not half a second.

TuxVinyards:
> It shouldn't. Exactly the same requests are being made either way

That's what I expected too.  I only ran with 'time' to see how long it took generally.

I randomly selected to try  '--check' on Bodhi the other day and was shocked at how slow it was. Raised the whole thing as a general issue ... #1108

The thing is, @Jai-JAP 's tweaks _do_ make checks run faster. The results are there in the screen shot.

And the `tr '\n' ` bit, that was just a _for example_ snippet. It was just something that caught my eye as I quickly scanned through. 

It's was just my initial impressions.

lj3954:
> The thing is, @Jai-JAP 's tweaks _do_ make checks run faster. The results are there in the screen shot.

Run multiple trials of each. There should be zero difference. There is no actual difference in the code, variables are just declared differently. With the web_check function, the one change made in this PR is that the redirect URL variable is set directly from a function call, rather than first being set to an empty string. That might buy you a few nanoseconds.

Jai-JAP:
> > The thing is, @Jai-JAP 's tweaks _do_ make checks run faster. The results are there in the screen shot.
> 
> Run multiple trials of each. There should be zero difference. There is no actual difference in the code, variables are just declared differently. With the web_check function, the one change made in this PR is that the redirect URL variable is set directly from a function call, rather than first being set to an empty string. That might buy you a few nanoseconds.

There will be some difference for functions with `tr '\n' ' '`. But for other functions it should be exactly the same.

flexiondotorg:
Thank you for working on this. It looks like we've been doing similar work :smile: My [tests branch](https://github.com/quickemu-project/quickemu/tree/tests) includes many of these changes.

I've included a few of yours I didn't have and will push that in a little bit. But, I won't be seeking to merge this is our work is very similar, and some of the variable assignment changes you've made introduce more `shellcheck` errors, something I want to avoid as @zen0bit is proposing some PRs to add more check to CI.

@TuxVinyards I am not motivated to invest significant effort in optimising the speed of the test, I am more interested in accuracy and we can run the tests in CI :+1: That said, my branch can run a full iso check in a reasonable time.

PR #1116, opened 2024-04-27
urgent fix for sonoma and #1114

### LINKED ISSUE #1114: bug: macos sonoma hang during startup
I have installed quickget cloning from github

I downloaded sonoma with quickget with the command
quickget macos sonoma
see the video 
https://youtu.be/07JLZCy5zlo

then I installed sonoma with the command 
quickemu --vm macos-sonoma.conf
see the vidoes of the begin of installation end the end 
start : https://youtu.be/MpbozxjQACw
end : https://youtu.be/Z5H0FT6Karo

it seem to me that almost at the end of the installation ( second startup )  the vm rebooted  
so probably the installation in not complete ( you can view that in https://youtu.be/Z5H0FT6Karo )

when I try to boot i get a blank screen ( after i have chosen the disk of installation )  and after sometime it restart  
see the video 
https://youtu.be/MpbozxjQACw

my system is fedora 39 

immagine

the log is void 

immagine

what can i check ?
P.S: i have tried do install also big-sur monterey ventura with the same result 

only the installation of catalina is working ( but is too old for what i have to do )

thank you
for any response

TuxVinyards:
fix #1114

### DIFF
--- a/quickemu
+++ b/quickemu
@@ -528,8 +528,8 @@ function vm_boot() {
         # A CPU with AVX2 support is required for >= macOS Ventura
         case ${macos_release} in
         ventura|sonoma)
-            if check_cpu_flag sse4_1 && check_cpu_flag avx2; then
-                CPU="-cpu Broadwell-noTSX-IBRS,kvm=on,vendor=GenuineIntel,+sse,+sse4.1,+avx,+avx2,+hypervisor,+popcnt,+aes,+xsave,+xsavec,+xsaveopt,+xgetbv1,+bmi1,+bmi2,+smep,+fma,+movbe,+invtsc"
+            if check_cpu_flag sse4_2 && check_cpu_flag avx2; then
+                CPU="-cpu Haswell-noTSX-IBRS,kvm=on,vendor=GenuineIntel,+sse3,+sse4.2,+aes,+xsave,+avx,+xsaveopt,+xsavec,+xgetbv1,+avx2,+bmi2,+smep,+bmi1,+fma,+movbe,+invtsc"
             else
                 echo "ERROR! macOS ${macos_release} requires a CPU with SSE 4.1 and AVX2 support."
                 exit 1


lj3954:
This doesn't fix that issue, since they experienced issues with Monterey and BigSur, both of which use the older Penryn CPU. That issue still needs to be addressed. This only fixes the issue I found, where the emulated Broadwell CPU doesn't work on AMD systems.

TuxVinyards:
@lj3954  Penryn is another issue.  I did flag up several months ago that Penryn was basically deprecated but you didn't want to know ....

Maybe  you approve this one and then we can solve the Penryn issue separately?

lj3954:
> @lj3954 Penryn is another issue. I did flag up several months ago that Penryn was basically deprecated but you didn't want to know ....
> 
> Maybe you approve this one and then we can solve the Penryn issue separately?

The Penryn CPU is old and it may be a good idea to update the argument. However, that's not relevant to issue #1114. The person reporting that issue experienced the same problem on both newer versions of macOS with the broadwell argument, and older ones with the Penryn argument. They also posted their lscpu 10 minutes after you responded here, showing that they have an Intel laptop CPU, not AMD. This PR does fix an issue, but the issue it fixes is not issue #1114.

lj3954:
Why is AVX2 removed from this argument? The flag is still being checked for, and I figured it was a requirement of the newer macOS releases.

TuxVinyards:
> Why is AVX2 removed from this argument? The flag is still being checked for, and I figured it was a requirement of the newer macOS releases.

It was in there twice ...

lj3954:
My mistake, you just moved it.

TuxVinyards:
This fix also makes things work for Intel.  We needed sse.4.2 on intel.  Not 4.1 as in the notes in the quickget script.

I am not that Penryn is the root cause of our woes.

Looking at the link given in quickget this morning  <https://www.nicksherlock.com/2020/06/installing-macos-big-sur-on-proxmox/>  Nick Sherlock is saying 4.2 even for Big Sur.   But we are using 4.1 as an else.

Also looking at  <https://github.com/search?q=repo%3Akholia%2FOSX-KVM%20%20GenuineIntel&type=code>  Kholia is using Penryn exclusively.

Plus note Nick Sherlock on Mac OS 13 says "On AMD you may find that the installer reboots while loading, or the loading bar doesn’t appear, if so replace ` “-cpu host” with “-cpu Haswell-noTSX” ` "  The bit that I added in ....

<https://www.nicksherlock.com/2022/10/installing-macos-13-ventura-on-proxmox/>

flexiondotorg:
Duplicated in #1117

PR #1117, opened 2024-04-28
Older mac fix - improvements

TuxVinyards:
TEST on OLDER RYZEN and notebooks please

Same Lines as already tested as working by @lj3954  but with revised greps  based on further research.

See added links  and notes.  Also see comments to #1116

### DIFF
--- a/quickemu
+++ b/quickemu
@@ -523,23 +523,38 @@ function vm_boot() {
         esac
         ;;
     macos)
-        #https://www.nicksherlock.com/2020/06/installing-macos-big-sur-on-proxmox/
+        # https://www.nicksherlock.com/2020/04/installing-macos-catalina-on-proxmox-with-opencore/
+        # https://www.nicksherlock.com/2022/10/installing-macos-13-ventura-on-proxmox/
+        # Penryn  https://github.com/search?q=repo%3Akholia%2FOSX-KVM%20%20GenuineIntel&type=code
+        # https://en.wikipedia.org/wiki/MacOS_version_history#Releases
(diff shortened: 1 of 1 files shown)


TuxVinyards:
Something else to note is that Penryn is only a two core baseline CPU model dating from 2007.  Even Haswell is getting long in the tooth at 2014.  <https://qemu.readthedocs.io/en/v9.0.0/system/qemu-cpu-models.html#preferred-cpu-models-for-intel-x86-hosts>

It might be worth considering Catalina plus to run on Haswell.  Same Ryzen tested formula as used for Ventura but with basic avx, not avx2.

lj3954:
> Penryn is only a two core baseline CPU model dating from 2007

Is there any real performance or usability difference caused by emulating an older CPU?

TuxVinyards:
I would imagine in the same way that a newer physical CPU has extra instruction sets and has more cores ... 

And that Nick Sherlock seems to use the newer models where possible. 

There's somewhere that I read in Qemu docs that Penryn should only be used for legacy emulation too.

lj3954:
> extra instruction sets

Are they being made use of in this case? macOS clearly is able to boot with the penryn argument; missing instructions should make the system fail to boot or crash regularly. More instruction sets is the entire reason why compatibility is reduced when emulating newer CPUs, and therefore only the bare minimum that fully supports the guest OS should be used.

> has more cores

The amount of cores has nothing to do with what CPU you emulate. You can emulate a 486 with 32 cores if you really want to.

flexiondotorg:
Duplicated in #1119

PR #1119, opened 2024-04-28
Catalina Haswell MacOS variant

TuxVinyards:
As #1117  but beef up the virtual CPU a tad ....

Something else to note is that Penryn is only a two core baseline CPU model dating from 2007.  Even Haswell is getting long in the tooth at 2014.  <https://qemu.readthedocs.io/en/v9.0.0/system/qemu-cpu-models.html#preferred-cpu-models-for-intel-x86-hosts>

It might be worth considering Catalina plus to run on Haswell.  Same Ryzen tested formula as used for Ventura but with basic avx, not avx2.


flexiondotorg (approved):
Initial testing on Catalina and Sonoma looks good. We might need to run through some more tests on all the supported macOS releases, but I'll merge this now, and we can address any issue we might discover in the future.

PR #1122, opened 2024-04-29
Tuxedo OS consistency

TuxVinyards:
#1109  was also about fixing half of the entries being in one format and half of them being in another

As per  decision  make these entries all  as 'tuxedo-os'


zen0bit:
Changed to tuxedo-os again in my PR..
So this could be closed...?

PR #1120, opened 2024-04-28
Refactor and updates to get most of the tests working

flexiondotorg:
The pull request refactors and updates various functions related to listing ISO URLs and checking if the ISO download URLs are valid.

You can now:

List all URLs for an OS via:

`quickget --url-all ubuntu`

Check all URLs for an OS via:

`quickget --check-all ubuntu`

It is also possible to run these tests for all support OS' via:

`quickget --url-all`
`quickget --check-all`

Here are the failed test from `--check-all`

```
FAIL: arcolinux-v24.05.11-large           https://mirror.accum.se/mirror/arcolinux.info/iso/v24.05.11/arcolinuxl-v24.05.11-x86_64.iso
FAIL: arcolinux-v24.05.11-small           https://mirror.accum.se/mirror/arcolinux.info/iso/v24.05.11/arcolinuxs-v24.05.11-x86_64.iso
FAIL: arcolinux-v24.02.01-large           https://mirror.accum.se/mirror/arcolinux.info/iso/v24.02.01/arcolinuxl-v24.02.01-x86_64.iso
FAIL: arcolinux-v24.02.01-small           https://mirror.accum.se/mirror/arcolinux.info/iso/v24.02.01/arcolinuxs-v24.02.01-x86_64.iso
FAIL: blendos-v3-gnome                    https://mirror.ico277.xyz/blendos/gnome/blendos-v3-stable-gnome.iso
FAIL: easyos-5.7                          https://distro.ibiblio.org/easyos/amd64/releases/kirkstone/2024/5.7/easy-5.7-amd64.img
FAIL: easyos-5.6.7                        https://distro.ibiblio.org/easyos/amd64/releases/kirkstone/2024/5.6.7/easy-5.6.7-amd64.img
FAIL: easyos-5.6.6                        https://distro.ibiblio.org/easyos/amd64/releases/kirkstone/2024/5.6.6/easy-5.6.6-amd64.img
FAIL: easyos-5.6.5                        https://distro.ibiblio.org/easyos/amd64/releases/kirkstone/2023/5.6.5/easy-5.6.5-amd64.img
FAIL: easyos-5.6.4                        https://distro.ibiblio.org/easyos/amd64/releases/kirkstone/2023/5.6.4/easy-5.6.4-amd64.img
FAIL: easyos-5.6.3                        https://distro.ibiblio.org/easyos/amd64/releases/kirkstone/2023/5.6.3/easy-5.6.3-amd64.img
[...]

PR #1125, opened 2024-04-30
Add support for Fedora 40

yeurch:
This change adds support for Fedora 40.

Downloads of all flavors have been tested, so there are no exceptions to add to `handle_missing()`.

### DIFF
--- a/quickget
+++ b/quickget
@@ -771,7 +771,7 @@ function editions_endless() {
 }
 
 function releases_fedora() {
-    echo 39 38
+    echo 40 39 38
 }
 
 function editions_fedora() {


flexiondotorg:
Thanks for the contribution :bow:  I've include Fedora 40 in this PR: https://github.com/quickemu-project/quickemu/pull/1120

It also make the Fedora configuration dynamic :+1:

yeurch:
> It also make the Fedora configuration dynamic 👍

This looks great!  Thanks @flexiondotorg.

It's a little curious that Fedora 38 is no longer listed in Fedora's own [JSON file](https://fedoraproject.org/releases.json), as it's officially supported [until next month](https://fedorapeople.org/groups/schedule/f-38/f-38-key-tasks.html), but I guess that means that `quickget` calls to download 38 have been broken for a while anyway, as the same JSON file is used to source the download URLs.  This is clearly not a QuickEMU issue though.

PR #1127, opened 2024-04-30
ci: Add PR template

zen0bit:


zen0bit:
As could be seen from Files changed

I wanna create PR template which will automatically trigger desired CI action
Depending on PR template filled by user creating PR.

Current idea:
If quickget was changed = quickget shellcheck
If added OS = List all URLs and Check new OS
If OS changed = List all URLs and Check that OS
If quickemu was changed = quickemu shellcheck
...

Then we can check only what we want, not everything and waste of resources...

CI script with logic will be added later...

flexiondotorg:
I'm going to close this.

Pull requests should be submitted:
  - as drafts, indicating that they are a work in progress or require discussion
  - as ready to merge, indicating they have been tested and considered complete and ready merging.

PR #1133, opened 2024-05-02
Make wget2 preferred downloader if found.

Jai-JAP:
Make wget2 the default downloader. (Is generally faster than the other two for downloads) 

 - Fall back to `wget` and then `curl` if not found (wget is faster for downloads than curl)
 - Add user agents to `wget`/`wget2`

### DIFF
--- a/quickget
+++ b/quickget
@@ -1220,7 +1220,7 @@ function web_get() {
     fi
 
     if command -v wget2 &>/dev/null; then
-        if ! wget2 --quiet --continue --tries=3 --read-timeout=10 --force-progress --progress=bar:force:noscroll "${URL}" -O "${DIR}/${FILE}" --user-agent "${USER_AGENT}" "${HEADERS[@]}"; then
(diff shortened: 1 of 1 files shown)


flexiondotorg:
Sorry, I am not going to merge this. 

I recently cleaned up the `quickget` to only use `curl` for all web capture/download operations because the code was littered with `wget`, wget2` and `curl`.

- With `wget` being linked to `wget2` on some distributions, we can't rely on the expected behaviour of calling `wget`. 
- `aria2` operation differs across distributions depending on what TLS/SSL library is was linked against.
- `curl` does present a consistent interface and predictable behaviour.

Prior to the release of quickemu 4.9.3 I will be removing the last two references to `wget` and `wget2` in `quickget` so that the only way to process web capture and download within `quickget` will be via `web_pipe()`, `web_check()`, `web_redirect()` and `web_get()`.

See this pull request for more details:

- https://github.com/quickemu-project/quickemu/pull/1087

PR #1144, opened 2024-05-03
Add error_unsupported_combination function

zen0bit:


flexiondotorg:
Sorry, I won't be merging this. It adds a single-use function that doesn't enhance the code readability or ease of maintainence.

zen0bit:
I am counted with use of this function also to stop checking for missing editions (fedora, arco etc.) in checks later...

Never mind...

PR #1145, opened 2024-05-03
Make ubuntu variants with dashes (not xubuntu, kubuntu etc..)

zen0bit:
We could change this and leave `ubuntustudio` same as was.

But it's the **only** one which deserve different naming. Because isn't just same ubuntu with different DE suite.

In my opinion...

I would like to see consistency even in distro naming.

But thats not really needed change so up to maintainer...


flexiondotorg:
Sorry, I will not be merging this. These ubuntu names are the name that are used within the Canonical build infrastructure for the official flavours.

The origins of Quickemu are for Ubuntu developers and flavour maintainers to quickly test new builds of the operating system. As such, each of the Ubuntu flavour teams expects to be able to install and run test VMs via their well-known names from within the Ubuntu project.

PR #1150, opened 2024-05-03
Add os_info to list_csv

zen0bit:


flexiondotorg:
I've included OS info in the prompts that inform the user a release and/or edition is required, putting this information in front of the users more often.

PR #1151, opened 2024-05-03
Add PRETTY_NAME variable

zen0bit:


flexiondotorg:
This approach isn't going to work as PRETTY_NAME is dynamically requested. The existing function works well.

PR #1154, opened 2024-05-03
fix: CI bug

zen0bit:
Will not work without this...


zen0bit:
Your current CI tests will fail due to this @flexiondotorg 
And you canceled testing of my working one...

zen0bit:
Still one flaw there...

Issue #1163, opened 2024-05-05
feat: add system info capability to quickemu to out useful information when reporting an issue


flexiondotorg:
Add a feature to quickemu to out useful system information to request when reporting issues. Examples of useful information are:

- OS
- CPU and CPU capabilities such as as SSE and AVX
- RAM
- GPUs
- Linux kernel
- `qemu` version
- `curl` version

TuxVinyards:
Good idea

flexiondotorg:
I've also updated the issue template: https://github.com/quickemu-project/quickemu/commit/65f065a327807ab2d4e3c6baec66b66065ed8b96

PR #1171, opened 2024-05-08
feat: add diagnostic output

philclifford:
A starter for 10 towards providing some system information output per #1163
A separate function in a script for now to facilitate testing, tweaking and discussion.


flexiondotorg:
Thanks for this, I used it as inspiration for `quickreport` and also included a shell bundle in the issue template based on this.

PR #1225, opened 2024-05-14
feat: add support for running quickemu on macOS hosts

flexiondotorg:
# Support for macOS hosts

This pull request makes Quickemu compatible with macOS hosts, meaning that `quickemu` can now be used to create and run virtual machines on macOS hosts.

Intel and Apple Silicon CPUs on the host are supported. However, Quickemu currently only has x86 OS images available, so Apple Silicon Macs are currently provisioned with x86 virtual machines. This will be addressed in the future.

# Improved compatibility for macOS guests

In addition, this pull request significantly improves the compatibility and performance when running macOS guest VMs, particularly if your host has an Intel CPU. Thanks to @TuxVinyards and @lj3954 for their research in this area. 

- Close #447
- Close #784
- Close #1236

PR #1241, opened 2024-05-18
refactor: flag tweaks

TuxVinyards:
as per comment https://github.com/quickemu-project/quickemu/pull/1225#issuecomment-2117555805


TuxVinyards:
Delete.  Just seen this:

```
        if [ "${HOST_CPU_VENDOR}" == "GenuineIntel" ]; then
            if ! check_cpu_flag vmx; then
```

BUT  the amd flag still needs looking at

PR #1247, opened 2024-05-23
fix: resolve portability issues between Linux distros and macOS

flexiondotorg:
# Portability

When running `quickemu`, `quickget` and `quickreport` across different Linux distros and macOS, there are several portability issues with how `tail`, `head` and `grep` are invoked. 

This pull request addresses those issues. 

# Compatibility

`quickreport` is also updated to work on macOS x86 and Apple Silicon.

# CI timeouts and runtime

Lastly, some CI jobs were getting wedged and consuming 360 minutes of runtime before being forcibly terminated. To address that, `web_check()` now has timeouts and retries, EOL Ubuntu releases are removed, Debian releases are limited to the latest release in a series, and some other distros cap the number of available releases.

### DIFF
--- a/quickget
+++ b/quickget
@@ -906,7 +906,7 @@ function releases_nitrux() {
 }
 
 function releases_nixos() {
-    echo unstable 23.11 23.05 22.11 22.05 21.11 21.05
+    echo unstable 23.11 23.05
 }
 
 function editions_nixos() {
--- a/quickget
+++ b/quickget
@@ -719,7 +719,7 @@ function editions_debian() {
 
 function releases_deepin() {
     #shellcheck disable=SC2046,SC2005
-    echo $(web_pipe "https://cdimage.deepin.com/releases/" | grep "href=" | cut -d'"' -f2 | grep -v "\.\." | grep -v nightly | grep -v preview | sed 's|/||g' | sort -r)
+    echo $(web_pipe "https://cdimage.deepin.com/releases/" | grep "href=" | cut -d'"' -f2 | grep -v "\.\." | grep -v nightly | grep -v preview | sed 's|/||g' | tail -n 10 | sort -r)
 }
 
 function releases_devuan() {
--- a/quickget
+++ b/quickget
@@ -697,12 +697,20 @@ function releases_crunchbang++() {
 }
 
 function releases_debian() {
+    local ARCHIVE=""
+    local MAJ=""
     local NEW=""
     local OLD=""
     NEW=$(web_pipe "https://cdimage.debian.org/debian-cd/" | grep '\.[0-9]/' | cut -d'>' -f 9 | cut -d'/' -f 1)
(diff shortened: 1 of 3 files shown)


quickget
```diff
@@ -697,12 +697,18 @@ function releases_crunchbang++() {
...
-    OLD=$(web_pipe "https://cdimage.debian.org/cdimage/archive/" | grep -e '>[1-9][0-9]\.' | grep -v 'live' | grep -v NEVER | cut -d'>' -f 9 | cut -d'/' -f 1 | tac | head -20)
-    #shellcheck disable=SC2086
-    echo ${NEW} ${OLD}
+    echo -n "${NEW}"
+    MAJ=$(echo "${NEW}" | cut -d'.' -f 1)
+    for i in {1..2}; do
+        CUR=$((MAJ - i))
+        OLD=$(web_pipe "https://cdimage.debian.org/cdimage/archive/" | grep folder | grep -v 'live' | grep -v NEVER | cut -d'"' -f 6 | grep ^"${CUR}" | tail -n 1 | tr -d '/')
+        echo -n " ${OLD}"
+    done
```
lj3954, line 710: Why are we making a request to the same URL twice? Instead, shouldn't you store that as a variable and then filter through the data in a loop.

```suggestion
    local OLD_DATA
    OLD_DATA="$(web_pipe "https://cdimage.debian.org/cdimage/archive/" | grep folder | grep -v 'live' | grep -v NEVER | cut -d'"' -f 6)"
    for i in {1..2}; do
        CUR=$((MAJ - i))
        OLD=$(grep ^"${CUR}"  <<< "${OLD_DATA}" | tail -n 1 | tr -d '/')
        echo -n " ${OLD}"
    done
```

quickget
```diff
@@ -2225,7 +2228,7 @@ function get_nwg-shell() {
     local HASH=""
     local ISO="nwg-live-${RELEASE}-x86_64.iso"
     local URL="https://sourceforge.net/projects/nwg-iso/files"
-    HASH="$(web_pipe "https://sourceforge.net/projects/nwg-iso/rss?path=/" | grep "${ISO}" | cut -d'>' -f3 | cut -d'<' -f1 | tail -1)"
+    HASH="$(web_pipe "https://sourceforge.net/projects/nwg-iso/rss?path=/" | grep "${ISO}" | cut -d'>' -f3 | cut -d'<' -f1 | tail -n -1)"
```
philclifford, line 2231: Still works but I think this extra '-' is a typo: 

```suggestion
    HASH="$(web_pipe "https://sourceforge.net/projects/nwg-iso/rss?path=/" | grep "${ISO}" | cut -d'>' -f3 | cut -d'<' -f1 | tail -n 1)"
```

philclifford (approved):
A small typo to fix I think, but otherwise LGTM.

Performance of --url with *ubuntu* compared  with most others is (for me this morning...) pretty dire.  Could well be cdimage.ubuntu.com but might be worth keeping an eye on.

quickget
flexiondotorg: Good idea. I've pushed the change.

quickget
flexiondotorg: Good catch. Correction made.

PR #1264, opened 2024-06-02
fix: don't declare SPICE as local within configure_ports. close #1240

mjoerg:
Functionality was split across several functions in e29173a16d5fb1546a725a28bc8973007b228bbe, so SPICE cannot be local anymore as it is also referenced in vm_boot.

PR #1268, opened 2024-06-05
fix: running quickemu in zsh and spice display

SafaSafari:
because of auto completion of zsh, piping to grep [0-9] not work
this patch will simply fix this problem by quoting [0-9]


SafaSafari:
oh
i saw spice display not working
who defined SPICE variable local and used it outside of function? =))

ilmari-lauhakangas:
Spice issue is fixed in #1264.

PR #1290, opened 2024-06-24
feat: implement underlying aarch64 compatibility

flexiondotorg:
This adds compatibility for running aarch64 on aarch64, but does not add support for aarch64 media detection/download.


flexiondotorg:
@cubic-dev-ai re-review this

PR #1302, opened 2024-06-24
fix: change holoiso regexp to match shorter release tags

### LINKED ISSUE #1301: bug: holoiso release tag no longer found
holoiso latest release tag is shorter so the regexp fails to match (so no release is found and prompted for and holoiso is missing from list_csv.

PR incoming

philclifford:
fixes #1301

### DIFF
--- a/quickget
+++ b/quickget
@@ -828,7 +828,7 @@ function editions_haiku() {
 }
 
 function releases_holoiso() {
-    web_pipe "https://github.com/HoloISO/releases/releases/latest" | grep -o -e 'releases/tag/[[:digit:]]\+\.[[:digit:]]\+\.[[:digit:]]' | head -n 1 | cut -d'/' -f 3
+    web_pipe "https://github.com/HoloISO/releases/releases/latest" | grep -o -e 'releases/tag/[[:digit:]]\+\(\.[[:digit:]]\+\)*' | head -n 1 | cut -d'/' -f 3
 }
 
 function releases_kali() {


quickget
```diff
@@ -828,7 +828,7 @@ function editions_haiku() {
 }
 
 function releases_holoiso() {
-    web_pipe "https://github.com/HoloISO/releases/releases/latest" | grep -o -e 'releases/tag/[[:digit:]]\+\.[[:digit:]]\+\.[[:digit:]]' | head -n 1 | cut -d'/' -f 3
+    web_pipe "https://github.com/HoloISO/releases/releases/latest" | grep -o -e 'releases/tag/[[:digit:]]\+\(\.[[:digit:]]\+\)*' | head -n 1 | cut -d'/' -f 3
```
lj3954, line 831: Most other releases_ and get_ functions rely on the GitHub API, which is almost certainly a better idea for performance & stability reasons. 
Additionally, the preferred style for quickget releases & editions functions appears to always involve use of echo, whether or not it's necessary.
```suggestion
    #shellcheck disable=SC2046,SC2005
    echo $(web_pipe "https://api.github.com/repos/HoloISO/releases/releases" | grep "tag_name" | cut -d'"' -f 4 | head -n +2)
```

lj3954 (changes requested):

quickget
lj3954: Releases should also never be truncated to the length 1, since in that case there's no point checking for releases, just offer the user a "latest" option and determine the latest release within the get_ function.

PR #1305, opened 2024-06-25
fix: Fixes nix flakes by adding version to package.nix

mystery3525:
Essentially, when you try and import the project directly with nix flakes, you receive a missing 'name' attribute error. This is because 'name' is supposed to be defined in stdenv.mkDerivation, but is not. I [looked into it](https://discourse.nixos.org/t/clarification-on-package-names-and-versions/9819/2) and found that `name` is actually `<pname>-<version>`, so I just added it.

Below are the before and after outputs of flake.nix using `nix flake show`:

### Before:
image

### After:
image

### DIFF
--- a/package.nix
+++ b/package.nix
@@ -52,11 +52,18 @@ let
     usbutils
     xdg-user-dirs
   ];
+  versionMatches =
+    builtins.match ''
+      .*
+      readonly[[:blank:]]VERSION="([[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+)"
+      .*
+    '' (builtins.readFile ./quickemu);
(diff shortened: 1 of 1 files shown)


mystery3525:
I was looking at how we did versioning and thought that having to update the version in two different places is too much work, so I discovered a neat approach by [facebook/hhvm](https://github.com/facebook/hhvm/blob/cebd9dd7818e7dfb6908f52e6812506942575d92/hhvm.nix#L76) where they dynamically pull the version from a file, and applied that here with `readonly VERSION="..."`.

flexiondotorg (approved):

Issue #916, opened 2024-01-31
quickget should detect or allow specify the host architecture when fetching iso


d7an:
#### Expected behaviour
I'm on Apple MacBook Pro with Apple Silicon M1 chip.  When running 'quickget' to fetch the iso, it doesn't detect the host architecture and attempt the fetch one that works on the host, and appear to favor x86_64/amd64 iso.

Supporting command line argument allowing user to specify the preferred architecture would also be a good option.

#### Actual behaviour

Seems to favor x86_64/amd architecture.

#### Steps to reproduce the behaviour

```
❯ arch   
arm64

❯ quickget --show-iso-url ubuntu 22.04
https://releases.ubuntu.com/22.04/ubuntu-22.04.3-desktop-amd64.iso

```

#### Quickemu output

See above.

#### Linux Distribution & Kernel

```bash
❯ uname -a
Darwin mbp-2021.local 23.3.0 Darwin Kernel Version 23.3.0: Wed Dec 20 21:30:44 PST 2023; root:xnu-10002.81.5~7/RELEASE_ARM64_T6000 arm64
```

lj3954:
There has been talk about architecture support for a while. I don't have capable ARM64 hardware, so someone else will have to do the work on quickemu. I've created a PR with ARM64 support in quickget for certain operating systems, but it'll have to wait until quickemu supports other architectures. #917

dabrown645:
Since qemu can emulate various architectures selecting the image should not be based solely on the current machines architecture. While that may be an appropriate default user should be able to select any supported architecture to get then have quickemu generate correct emulation environment for it.

lj3954:
> Since qemu can emulate various architectures selecting the image should not be based solely on the current machines architecture. While that may be an appropriate default user should be able to select any supported architecture to get then have quickemu generate correct emulation environment for it.

This was specifically part of my draft PR, although the way the original quickget creates a subshell to access the ISO URL caused it to be nearly impossible for the function to directly relay that. Instead, I had to check for strings like 'aarch64' in the filename, which is obviously not as reliable. Still, it leaves an "arch" value in the config file, but only if it's not amd64 (which in my eyes should undoubtedly remain as a default for quickemu).

PR #1321, opened 2024-06-27
feat: Add quickgetc

lj3954:
Quickgetc is an implementation of bash quickget which makes use of [configurations](https://github.com/lj3954/quickget_configs) generated through [quickget_ci](https://github.com/lj3954/quickemu-rs/tree/main/quickget_ci). This is very early, most features are not yet implemented.

Advantages:
- Quickgetc itself will require little maintenance or updates, since everything is handled in CI
- Releases and editions for all implemented operating systems are determined dynamically during the config generation, with no performance impact to the end user
- All URLs are validated, entries with invalid URLs are thrown out
- Quickgetc uses the host's architecture by default, falling back if it's not available for the selected distro. The user can also override it. 
- Archives are automatically handled, no need to play guessing games or implement features per OS
- list_csv is fast. Really fast. I've seen it under 20ms. Wonderful for projects like QuickGUI
- Complete compatibility with bash quickget and quickemu

Additional dependency: zstd. Alternatively, gzip could be used (which is already used by bash quickget) since the JSON is compressed into both zstd and gzip archives, but zstd is faster and has slightly better compression.

- [x] Handle file downloads
- [x] Create VM configuration
- [ ] Re-implement test functionality, create_vm, and other options from quickget
- [ ] Implement support for custom disk images, including those sourced from the web, which quickget_ci supports and already uses in one case (FreeBSD VM Images)
- [ ] Add far greater choice of operating systems to quickget_ci

### DIFF
--- a/quickgetc
+++ b/quickgetc
@@ -1,0 +1,657 @@
+#!/usr/bin/env bash
+
+CACHE_FILE="quickget_data.json.zst"
+CACHE_URL="https://github.com/lj3954/quickget_configs/releases/download/daily/quickget_data.json.zst"
+IMAGE_TYPES=("iso" "img" "fixed_iso" "floppy")
+QUICKGETC_VERSION="0.1.0"
+
+function cleanup() {
+    if [ -n "$(jobs -p)" ]; then
+        kill "$(jobs -p)" 2>/dev/null
+    fi
+}
+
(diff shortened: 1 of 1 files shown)


lj3954:
Closes #916

PR #1373, opened 2024-07-10
feat: Improve quickget custom configurations, including unattended windows support

### LINKED ISSUE #1369: feat: create unattended windows installations with --create-config
- [x] I have searched the issues and this feature has not previously been requested

I often want to create a number of windows virtual machines, but quickget always downloads the iso, even if i have the iso with me. So I tried the `--create-config` flag to create a windows 10 config file. But it is not an unattended install.

It would be really helpful if there is a way to use unattended install (with auto installation of virtio drivers, etc) with using pre-downloaded windows iso.

**Describe alternatives you've considered**
I've used quickget a couple of times to get ip rate-limited, and it skips the download step but creates an unattended installation iso with virtio drivers.

sudoAlphaX:
- **refactor: move create_config function to the bottom**
- **refactor: move unattended win installation code inside function**
- **feat: add --unattended flag to --create-config to create unattend.iso (windows hosts)**

# Description

Ability to pass in `--unattended` to `--create-config` to create unattended.iso and download VirtIO drivers for Windows guests.

- Closes #1369

## Type of change

# Checklist:

- [x] I have made corresponding changes to the documentation (*remove if no documentation changes were required*)

### DIFF
--- a/quickget
+++ b/quickget
@@ -3333,18 +3333,16 @@ function create_config() {
         *windows-server*|*eval_oemret_x*|*eval_x*) CUSTOM_OS="windows-server";;
         *windows*|win*)
             CUSTOM_OS="windows"
-            while [ $# -gt 2 ]; do
-                case "${3}" in
-                    -unattended|--unattended)
-                        echo "Downloading VirtIO drivers..."
-                        web_get "https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/stable-virtio/virtio-win.iso" "${VM_PATH}"
-                        FIXED_ISO=${VM_PATH}/virtio-win.iso
-
-                        rm -f "${VM_PATH}/unattended.iso"
-                        unattended_windows "${VM_PATH}"
-                        shift
-                esac
-            done
+            # Older windows 10 ISOs use the year followed by the month rather than the year & half). Match any text for language.
+            if [ "${3}" != "--disable-unattended" ] && ( [ "${3}" == "--unattended" ] || grep -E -q 'Win(10|11)_([0-9]{2}H(1|2)|[0-9]{4})_[^.]*?(x64|x32)(v[0-9])?.iso' <<< "${INPUT}" ); then
+                echo "Creating unattended Windows installation files. To disable, pass --disable-unattended"
+                echo
+                echo "Downloading VirtIO drivers..."
+                web_get "https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/stable-virtio/virtio-win.iso" "${VM_PATH}"
+                FIXED_ISO=${VM_PATH}/virtio-win.iso
+                rm -f "${VM_PATH}/unattended.iso"
+                unattended_windows "${VM_PATH}"
+            fi
             ;;
 
         *) CUSTOM_OS="linux";;
@@ -3397,7 +3395,7 @@ Arguments:
   --help                                   : Show this help message
 ------------------------------------ Flags -------------------------------------
 --create-config:
-  --unattended                             : Unattended windows installation
+  --disable-unattended                     : Force quickget not to set up an unattended installation
 -------------------------- For testing & development ---------------------------
   --url           [os] [release] [edition] : Show image URL(s)
   --check         [os] [release] [edition] : Check image URL(s)
--- a/quickget
+++ b/quickget
@@ -308,51 +308,6 @@ function csv_data() {
     wait
 }
 
-function create_config() {
-    local VM_PATH="${1}"
-    local INPUT="${2}"
-    OS="custom"
-    if ! mkdir "${VM_PATH}" 2>/dev/null; then
-        echo "ERROR! Could not create directory: ${VM_PATH}. Please verify that it does not already exist"
-        exit 1
-    fi
-    if [[ "${INPUT}" == "http://"* ]] || [[ "${INPUT}" == "https://"* ]]; then
-        INPUT="$(web_redirect "${INPUT}")"
-        if [[ "${INPUT}" == *".iso" ]] || [[ "${INPUT}" == *".img" ]]; then
-            web_get "${INPUT}" "${VM_PATH}"
-            INPUT="${VM_PATH}/${INPUT##*/}"
-        else
-            echo "ERROR! Only ISO,IMG and QCOW2 file types are supported for --create-config"
-            exit 1
-        fi
-    fi
-    if [[ "${INPUT}" == *".iso" ]]; then
-        echo "Moving image to VM dir" && mv "${INPUT}" "${VM_PATH}"
-        CUSTOM_IMAGE_TYPE="iso"
-    elif [[ "${INPUT}" == *".img" ]]; then
-        echo "Moving image to VM dir" && mv "${INPUT}" "${VM_PATH}"
-        CUSTOM_IMAGE_TYPE="img"
-    elif [[ "${INPUT}" == *".qcow2" ]]; then
-        echo "Moving image to VM dir" && mv "${INPUT}" "${VM_PATH}/disk.qcow2"
-        CUSTOM_IMAGE_TYPE="qcow2"
-    else
-        echo "ERROR! Only ISO,IMG and QCOW2 file types are supported for --create-config"
-        exit 1
-    fi
-    echo "Creating custom VM config for ${INPUT##*/}."
-    case "${INPUT,,}" in
-        *windows-server*) CUSTOM_OS="windows-server";;
-        *windows*) CUSTOM_OS="windows";;
-        *freebsd*) CUSTOM_OS="freebsd";;
-        *kolibrios*) CUSTOM_OS="kolibrios";;
-        *reactos*) CUSTOM_OS="reactos";;
-        *) CUSTOM_OS="linux";;
-    esac
-    echo "Selecting OS: ${CUSTOM_OS}. If this is incorrect, please modify the config file to include the correct OS."
-    echo
-    make_vm_config "${INPUT}"
-}
-
 function list_supported() {
     list_csv | cut -d ',' -f2,3,4 | tr ',' ' '
     exit 0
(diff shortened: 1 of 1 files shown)


quickget
```diff
@@ -3326,6 +3286,72 @@ function create_vm() {
...
+                        echo "Downloading VirtIO drivers..."
+                        web_get "https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/stable-virtio/virtio-win.iso" "${VM_PATH}"
+                        FIXED_ISO=${VM_PATH}/virtio-win.iso
+
+                        rm -f "${VM_PATH}/unattended.iso"
+                        unattended_windows "${VM_PATH}"
+                        shift
+                esac
+            done
+            ;;
```
lj3954, line 3339: First off, I have no idea why we're using a loop for this. There are no parameters that could possibly be passed here other than your proposed --unattended. In addition, it's an infinite loop if there's a third argument that's not --unattended, since shift is only executed if the value is --unattended.

I also disagree that we should require an extra parameter for this, unattended installation should be the default and automatically handled for the user if possible. Therefore, I propose using pattern matching to determine if the specified ISO has a name matching Microsoft's formatting for consumer Windows ISOs, with optional flags to override it

Please validate the code and formatting before including it, of course. It may not be completely correct. I've written it in the GitHub comment interface rather than within my editor.

```suggestion
            # After checking flags, determine whether ISO is named as a Windows 10 or 11 ISO using pattern matching
            # Older windows 10 ISOs use the year followed by the month rather than the year & half). Match any text for language.
            if [ "${3}" != "--disable-unattended" ] && ( [ "${3}" == "--unattended" ]
                || grep -E -q 'Win(10|11)_([0-9]{2}H(1|2)|[0-9]{4})_[^.]*?(x64|x32)(v[0-9])?.iso' <<< "${INPUT}" ); then
                echo "Downloading VirtIO drivers..."
                web_get "https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/stable-virtio/virtio-win.iso" "${VM_PATH}"
                FIXED_ISO=${VM_PATH}/virtio-win.iso

                rm -f "${VM_PATH}/unattended.iso"
                unattended_windows "${VM_PATH}"
            fi
```

quickget
```diff
@@ -3326,6 +3286,72 @@ function create_vm() {
...
+        *reactos*) CUSTOM_OS="reactos";;
+        *) CUSTOM_OS="linux";;
+    esac
+    echo "Selecting OS: ${CUSTOM_OS}. If this is incorrect, please modify the config file to include the correct OS."
+    echo
+    if [ -n "${FIXED_ISO}" ]; then
+        make_vm_config "${INPUT}" "${FIXED_ISO}"
+    else
+        make_vm_config "${INPUT}"
+    fi
```
lj3954, line 3352: The check for whether the fixed ISO parameter has a non-empty value is already present within make_vm_config. Remove the unnecessary check.
```suggestion
    make_vm_config "${INPUT}" "${FIXED_ISO}"
```

quickget
```diff
@@ -3326,6 +3286,72 @@ function create_vm() {
...
+        echo "Moving image to VM dir" && mv "${INPUT}" "${VM_PATH}/disk.qcow2"
+        CUSTOM_IMAGE_TYPE="qcow2"
+    else
+        echo "ERROR! Only ISO,IMG and QCOW2 file types are supported for --create-config"
+        exit 1
+    fi
+    echo "Creating custom VM config for ${INPUT##*/}."
+    case "${INPUT,,}" in
+        *windows-server*) CUSTOM_OS="windows-server";;
+        *windows*)
```
lj3954, line 3325: I wrote this, but it's practically worthless. Let's fix it to correctly match enterprise evaluation & consumer windows ISOs respectively. Let's also shift all other matches ahead of especially server, to avoid accidental matches of some of these patterns.
```suggestion
        *freebsd*) CUSTOM_OS="freebsd";;
        *kolibrios*) CUSTOM_OS="kolibrios";;
        *reactos*) CUSTOM_OS="reactos";;
        *windows*|win*)
        *windows-server*|*eval_oemret_x*|*eval_x*) CUSTOM_OS="windows-server";;
```

quickget
```diff
@@ -3326,6 +3286,72 @@ function create_vm() {
...
+                        rm -f "${VM_PATH}/unattended.iso"
+                        unattended_windows "${VM_PATH}"
+                        shift
+                esac
+            done
+            ;;
+
+        *freebsd*) CUSTOM_OS="freebsd";;
+        *kolibrios*) CUSTOM_OS="kolibrios";;
+        *reactos*) CUSTOM_OS="reactos";;
```
lj3954, line 3343:
```suggestion
```

quickget
```diff
@@ -3326,6 +3286,72 @@ function create_vm() {
...
+    OS="custom"
+    if ! mkdir "${VM_PATH}" 2>/dev/null; then
+        echo "ERROR! Could not create directory: ${VM_PATH}. Please verify that it does not already exist"
+        exit 1
+    fi
+    if [[ "${INPUT}" == "http://"* ]] || [[ "${INPUT}" == "https://"* ]]; then
+        INPUT="$(web_redirect "${INPUT}")"
+        if [[ "${INPUT}" == *".iso" ]] || [[ "${INPUT}" == *".img" ]]; then
+            web_get "${INPUT}" "${VM_PATH}"
+            INPUT="${VM_PATH}/${INPUT##*/}"
```
lj3954, line 3303: Unrelated to this PR, but this line is incorrect. Let's fix it. The VM path cannot be included within the input.
```suggestion
            INPUT="${INPUT##*/}"
```

quickget
```diff
@@ -3326,6 +3286,72 @@ function create_vm() {
...
+    elif [[ "${INPUT}" == *".img" ]]; then
+        echo "Moving image to VM dir" && mv "${INPUT}" "${VM_PATH}"
+        CUSTOM_IMAGE_TYPE="img"
+    elif [[ "${INPUT}" == *".qcow2" ]]; then
+        echo "Moving image to VM dir" && mv "${INPUT}" "${VM_PATH}/disk.qcow2"
+        CUSTOM_IMAGE_TYPE="qcow2"
+    else
+        echo "ERROR! Only ISO,IMG and QCOW2 file types are supported for --create-config"
+        exit 1
+    fi
```
lj3954, line 3321: In case the user points to a file outside the current directory.
```suggestion
    fi
    INPUT="$(basename "${INPUT}")"
```

quickget
```diff
@@ -3326,6 +3286,72 @@ function create_vm() {
...
+        INPUT="$(web_redirect "${INPUT}")"
+        if [[ "${INPUT}" == *".iso" ]] || [[ "${INPUT}" == *".img" ]]; then
+            web_get "${INPUT}" "${VM_PATH}"
+            INPUT="${VM_PATH}/${INPUT##*/}"
+        else
+            echo "ERROR! Only ISO,IMG and QCOW2 file types are supported for --create-config"
+            exit 1
+        fi
+    fi
+    if [[ "${INPUT}" == *".iso" ]]; then
```
lj3954, line 3309:
```suggestion
    if [ ! -f "${INPUT}" ]; then
        echo "ERROR! The input must be a valid URL or path to an ISO, IMG, or QCOW2 file."
        exit 1
    elif [[ "${INPUT}" == *".iso" ]]; then
```

quickget
```diff
@@ -3363,11 +3389,14 @@ Advanced usage:
...
   --download      <os> <release> [edition] : Download image; no VM configuration
-  --create-config <os> [path/url]          : Create VM config for a OS image
+  --create-config <os> [path/url] [flags]  : Create VM config for a OS image
   --open-homepage <os>                     : Open homepage for the OS
   --show          [os]                     : Show OS information
   --version                                : Show version
   --help                                   : Show this help message
+------------------------------------ Flags -------------------------------------
+--create-config:
+  --unattended                             : Unattended windows installation
```
lj3954, line 3399:
```suggestion
  --unattended                             : Force quickget to set up an unattended windows installation
  --disable-unattended               : Force quickget not to set up an unattended installation
```

lj3954 (changes requested):
I think some major changes need to be made. To the function in general, but also your proposed additions.

quickget
sudoAlphaX: > First off, I have no idea why we're using a loop for this.
The reason i used a loop is for extensibility. It will be easier to add additional flags for specific type of installations if required in the future.

> I also disagree that we should require an extra parameter for this, unattended installation should be the default and automatically handled for the user if possible. 

Thats absolutely correct. My proposed solution requires users to rename their iso to "windows.iso". Having pattern matching would simply be better.

lj3954: > My proposed solution requires users to rename their iso to "windows.iso". Having pattern matching would simply be better.

No, it doesn't. I'm not sure what you're talking about. Your solution requires users to pass "--unattended" after to set up unattended setup. I've already fixed the issue where quickget wouldn't detect Windows ISOs using standard naming in 5d64018, which I've committed directly to your branch due to it wasting a bunch of space in conversations for issues that don't have to do with your proposed changes.

See my suggested changes, If the flags --disable-unattended or --unattended aren't present, with my changes, quickget would check whether the ISO follows Microsoft's naming scheme for windows 10 or 11 ISOs using a regex. This would remove the need to pass --unattended as long as you don't rename the ISO. 

To explain the syntax used.
Win(10|11) - pretty self explanatory, matches Win and then 10 or 11
_([0-9]{2}H(1|2)|[0-9]{4}) - matches either 2 digits (year), H, and then 1 or 2 (halves of year) (e.g. 23H2), or 4 digits (year + month) for older Windows 10 ISOs (e.g. 1607)
_[^.]*? - matches an underscore and then any character other than a period or specific other characters such as newlines. This allows us to skip languages, which can be formatted in many ways, with multiple underscores.
(x64|x32) - self explanatory.  Matches either x64 or x32, Microsoft's naming for the x86_64 and i686 architectures
(v[0-9])? - optionally matches v followed by one digit, for versions of some Windows ISOs

sudoAlphaX: > No, it doesn't. I'm not sure what you're talking about.

I apologize to have quoted the wrong block of text. Let me correct that:

> Therefore, I propose using pattern matching to determine if the specified ISO has a name matching Microsoft's formatting for consumer Windows ISOs

Thats absolutely correct. My proposed solution requires users to rename their iso to "windows.iso" (which is not optimal). Having pattern matching would simply be better (which is your proposed solution).

> the point of quickemu is that you don't have to read a manual to set up QEMU

I agree with that. Lets use your solution then.

lj3954 (approved):

flexiondotorg (approved):

PR #1378, opened 2024-07-12
feat: allow skipping downloads using SIGINT

sudoAlphaX:
# Description

Downloads get skipped using SIGINT instead of terminating script

## Type of change

# Checklist:

### DIFF
--- a/quickget
+++ b/quickget
@@ -1201,6 +1201,9 @@ function web_pipe() {
 
 # Download a file from the web
 function web_get() {
+
+    trap 'echo && echo "Skipping download" && echo; trap '"'exit'"' SIGINT' SIGINT
+
     local CHECK=""
     local HEADERS=()
     local URL="${1}"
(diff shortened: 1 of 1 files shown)


lj3954:
The partially downloaded file will still be linked to in the VM configuration, and there's no reason I can see that you'd want that to happen. It's also nearly universally the case that SIGINT is used to exit programs, this would likely cause a nuisance for many. I oppose this change, but I'll request a second review in case the head of the project has a different opinion of it.

flexiondotorg:
What is the use case for this proposed change?

sudoAlphaX:
When I was trying to create a Windows VM using --create-config flag with the existing iso, quickemu automatically started downloading the virtio drivers and spice drivers (#1373). For some reason, the virtio drivers were downloading really slowly.
Hence, I thought it would be useful if we were able to skip the particular download, because I already had that driver iso.

PR #1410, opened 2024-08-04
fix: extend search for smbd

### LINKED ISSUE #1409: bug: smb support broken
- [*] I have searched the issues and this bug has not been reported previously 

...but it's caused by an attempt to fix (#754) another issue (#722)

**Describe the bug**

The samba demon will not be started, because it will not be found by the mechanism suggested in #722 / #754
**It would only work, if "/usr/sbin/" is included in your [search-]PATH**, but that's usually not the case for ordinary users.

**To Reproduce**

As long as this "Fix" doesn't get reverted the qemu-smb-entry will never be used on `debian GNU Linux` systems and similar setups.  

**Expected behavior**

The smb related qemu entry should be used, if samba is installed at the common location on the machine and the PUBLIC directory exists -- i.e. just the old behavior!

Please revert #754

philclifford:
# Description

add alternate test for /usr/sbin/smbd in case it is not in the user's PATH

- Closes #1409

## Type of change

# Checklist:

### DIFF
--- a/quickemu
+++ b/quickemu
@@ -1086,7 +1086,9 @@ function configure_file_sharing() {
         fi
 
         # SMB
-        if [ -x "$(command -v smbd)" ]; then
+        # We need to search in NixOS compatible paths as well as the standard location
+        # since /usr/sbin/smbd may not be in the PATH.
+        if [ -x "$(command -v smbd)" ] || [ -x "/usr/sbin/smbd" ]; then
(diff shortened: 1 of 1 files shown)


mash-graz:
Perhaps we should also add again the test for the existence of the `${PUBLIC}` folder in the condition chain.

flexiondotorg (approved):

PR #1468, opened 2024-10-05
fix:  Archcraft and  ReactOS ISO naming and editions quoting

TuxVinyards:
Otherwise just downloads with filename 'download' without even a .iso extension.

Also includes a fix for  a  shellcheck highlighted quote glitch.

- [x ] Bug fix (non-breaking change which fixes an issue)

- [ x] I have performed a self-review of my code

- [ x] I have tested my code in common scenarios and confirm there are no regressions


quickget
```diff
@@ -1547,6 +1547,7 @@ function get_archcraft() {
     local HASH=""
     local URL=""
     URL="https://sourceforge.net/projects/archcraft/files/${RELEASE}/download"
+    URL="$(curl -sLIf -m 30 --retry 3 "${URL}" | grep -i 'location:' | grep -v mirror | cut -d: -f2- | cut -d? -f1)"
```
lj3954, line 1550: Could we not use the web_get function instead?

lj3954: Sorry, web_redirect is probably the right function instead.

TuxVinyards: @lj3954 

Fair enough,  web_redirect would be more consistent.   I only like to spend a limited time looking the vanilla quickget code these days.  

I going to leave this as a 'Heads Up'.  Please feel free to close things.  

When I tested the web_redirect method with Archcraft, I personally couldn't get it to work:

```
#!/usr/bin/quickemu --vm
guest_os="linux"
disk_img="archcraft-latest/disk.qcow2"
iso="archcraft-latest/archcraft-2024.07.01-x86_64.iso?viasf=1"
```

Changing the URL line to mimic that used by Zorin, as in `URL=$(web_redirect "https://sourceforge.net/projects/archcraft/files/${RELEASE}/download")`  still  leaves me needing to add a pipe to `cut -d? -f1`  or similar, at some point. 

Looking closely, I  notice that the quickget logic actually gets routed far more than once through web_get and web_redirect on its journey.  Maybe you can see where Zorin manages to get the tail end instructions removed?

It was just a couple of one-liner fixes that probably some misguided reason I thought  that I would pass on ....

lj3954 (approved):

PR #1479, opened 2024-10-15
fix: dynamic grep latest MX linux and sort Kolibri

TuxVinyards:
# Description

Fixes:

FAIL: mxlinux-23.3-Xfce                   https://sourceforge.net/projects/mx-linux/files/Final/Xfce/MX-23.3_x64.iso

FAIL: mxlinux-23.3-KDE                    https://sourceforge.net/projects/mx-linux/files/Final/KDE/MX-23.3_KDE_x64.iso

FAIL: mxlinux-23.3-Fluxbox                https://sourceforge.net/projects/mx-linux/files/Final/Fluxbox/MX-23.3_fluxbox_x64.iso

## Type of change

Hopefully 'maintenance free'

- [ x] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [ x] I have performed a self-review of my code

- [ x] I have tested my code in common scenarios and confirmed there are no regressions

- [ x] I have added comments to my code, particularly in hard-to-understand sections


philclifford (changes requested):
There are commits mixed in that make changes to KolibriOS. Please refactor these into a seperate PR for KolibriOS changes and limit this one to just the MX changes referred to in the title.
And thanks for catching and fixing up KolibriOS btw !

quickget
```diff
@@ -823,7 +823,8 @@ function releases_manjaro() {
 }
 
 function releases_mxlinux() {
-    echo 23.3
+    # needs header, so not web_pipe:
+    curl -Ils "https://sourceforge.net/projects/mx-linux/files/latest/download" | grep -i -m 1 'location' | cut -d? -f1 | cut -d_ -f1 | cut -d- -f3
```
philclifford, line 831: would it be better to pin the result more tightly with '^location:' ? This probably removes the need for ' -m 1 ' too but it can stay for me - harmless and protective.

TuxVinyards: :wave:   I did split this PR up from the Archcraft fix #1468 and tried to keep them more separate.  But then merge checker kept telling that it wouldn't accept the code unless I made another commit. 

See comments to dba4ce4  > Amend the commit message to match the pull request title, or *add another commit.*

There's  a whole load of distro 'FAILS' flagged up by the quickget checker that need attention far more.   And maybe the merge checker as well ...  

Something I would to see too, are some guideline notes added to the PR template to take the guess work out of what exact **title categories**  are permissible, as the merge checker gets nit-picking on that too.  

```
FAIL: fedora-39-Onyx                      
FAIL: fedora-39-Sericea                   
FAIL: holoiso-latest                      
FAIL: kolibrios-latest                    https://builds.kolibrios.org/eng/kolibri.iso
FAIL: mxlinux-23.3-Xfce                   https://sourceforge.net/projects/mx-linux/files/Final/Xfce/MX-23.3_x64.iso
FAIL: mxlinux-23.3-KDE                    https://sourceforge.net/projects/mx-linux/files/Final/KDE/MX-23.3_KDE_x64.iso
FAIL: mxlinux-23.3-Fluxbox                https://sourceforge.net/projects/mx-linux/files/Final/Fluxbox/MX-23.3_fluxbox_x64.iso
FAIL: opensuse-16.0                       https://download.opensuse.org/distribution/leap/16.0/iso/openSUSE-Leap-16.0-DVD-x86_64-Current.iso
FAIL: rebornos-latest                     
FAIL: truenas-scale-23                    
FAIL: ubuntu-server-23.10                 https://old-releases.ubuntu.com/releases/23.10/
```

It's not just these though.   The Github summarizer just greps for 'FAIL'.   There are quite a few fails that have text only wording that I noticed, eg 

```
PASS: ubuntustudio-24.10                  https://cdimage.ubuntu.com/ubuntustudio/releases/24.10/release/ubuntustudio-24.10-desktop-amd64.iso
Ubuntu Studio dvd is currently unavailable. Please select other OS/Release combination
PASS: vanillaos-22.10-r8                  https://github.com/Vanilla-OS/live-iso/releases/download/22.10-r8/VanillaOS-22.10-all.20230226.iso
```

If  a reverse grep was used, eg ` grep -v PASS | grep -v SKIP` then the text only errors would show up.

I have just added your BlendOS fix BTW  :+1:

philclifford:
See https://www.conventionalcommits.org/en/v1.0.0/ 
The merge checker's nagging is just to enforce "a single commit message must match the PR title exactly OR there must be several commits to merge".  The PR should contain atomic commits that achieve the intent of the PR - ideally a single conventional commit. 
It is preferable to keep the scope of the PR narrow - a single fix or feature. This way the resulting changelog is clear and straighforward and also the scope of changes that may need to be reviewed, tested and possibly ultimately rolled back or re-visited is kept small and coherent.

quickget
philclifford: There seem to be some issues with the `quickget --check` behaviour: for example the batch mode for SliTaz produces a false negative without FAIL: all the time whereas a direct `quickget --check slitaz loram` mostly gets a PASS: (although sometimes it results in the non-FAIL fail).  I suspect it may be some sort of rate-limiting behaviour at a mirror or CDN as it is usually the "last" option and it "gets better" with no apparent change at the far end (i.e. it is not in-flight updates to remote files)

quickget
```diff
@@ -3284,9 +3289,19 @@ function create_vm() {
...
+                    7z e "${VM_PATH}/${ISO}" -o"${VM_PATH}" >/dev/null 2>&1
+                    rm -f "${VM_PATH}/${ISO}"
+                    ISO="$(ls -1 "${VM_PATH}/"*.iso)"
+                    ISO="$(basename "${ISO}")"
+                fi
+            fi;;
         reactos)
             if [[ ${ISO} = *".zip"* ]]; then
-                unzip -qo "${VM_PATH}/${ISO}" -d "${VM_PATH}"
+                unzip -qo "${VM_PATH}/${ISO}" -d "${VM_PATH}"
```
philclifford, line 3304:
```suggestion
                unzip -qo "${VM_PATH}/${ISO}" -d "${VM_PATH}"
```

philclifford (approved):

PR #1487, opened 2024-10-18
fix(deepin):  Correct architecture insertion logic

### LINKED ISSUE #1488: bug: Deepin Release 23 quickget fails to find iso
It is not looking in the ${ARCH} subdirectory as intended.

philclifford:
so Deepin 23 and  later are correctly adjusted.

- Resolves #1488 
# Checklist:

### DIFF
--- a/quickget
+++ b/quickget
@@ -1694,7 +1694,7 @@ function get_deepin() {
     # deepin-desktop-community-20.3-amd64.iso
     local URL="https://cdimage.deepin.com/releases/"${RELEASE}
     # Correct URL for 23-RC onwards which has architecture directories
-    if [[ ! "${RELEASE}" == "20.9" ]] ; then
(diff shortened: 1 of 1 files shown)


quickget
```diff
@@ -1694,7 +1694,7 @@ function get_deepin() {
     # deepin-desktop-community-20.3-amd64.iso
     local URL="https://cdimage.deepin.com/releases/"${RELEASE}
     # Correct URL for 23-RC onwards which has architecture directories
-    if [[ "${RELEASE}" == *"23-RC"* ]] ; then
+    if [[ ! "${RELEASE}" == "20.9" ]] ; then
```
lj3954, line 1697:
```suggestion
    if [ "${RELEASE}" != "20.9" ] ; then
```

lj3954 (approved):

PR #1496, opened 2024-10-27
fix(solus): update Solus URLs

silkeh:
# Description

The latest Solus release has changed the naming of ISOs:

- Update all related functions to support the ISO name format used in the latest release.
- Use the directory listings in order to support more releases in the future.

## Type of change

# Checklist:

### DIFF
--- a/quickget
+++ b/quickget
@@ -982,12 +982,12 @@ function releases_slitaz() {
 
 function releases_solus() {
     #shellcheck disable=SC2046,SC2005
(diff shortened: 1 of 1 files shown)


lj3954:
I think it may be a better idea to allow web_pipe to take in extra arguments for curl, rather than creating a separate function for a single header to be passed.

philclifford (approved):

flexiondotorg (approved):

PR #1516, opened 2024-11-13
feat(quickget): Show download path

abitrolly:
# Description

I want to know where big files are stored without opening the browser.

## Type of change

# Checklist:

- [ ] I have made corresponding changes to the documentation (*remove if no documentation changes were required*)

### DIFF
--- a/quickget
+++ b/quickget
@@ -1218,6 +1218,7 @@ function web_get() {
     if [[ ${OS} != windows && ${OS} != macos && ${OS} != windows-server ]]; then
         echo "Downloading $(pretty_name "${OS}") ${RELEASE} ${EDITION}"
         echo "- URL: ${URL}"
+        echo "- PATH: ${DIR}/${FILE}"
     fi
 
     if ! curl --progress-bar --location --output "${DIR}/${FILE}" --continue-at - --user-agent "${USER_AGENT}" "${HEADERS[@]}" -- "${URL}"; then
--- a/quickget
+++ b/quickget
@@ -1218,7 +1218,7 @@ function web_get() {
     if [[ ${OS} != windows && ${OS} != macos && ${OS} != windows-server ]]; then
         echo "Downloading $(pretty_name "${OS}") ${RELEASE} ${EDITION}"
         echo "- URL: ${URL}"
-        echo "- PATH: ${DIR}/${FILE}"
+        echo "- PATH: ${PWD}/${DIR}/${FILE}"
     fi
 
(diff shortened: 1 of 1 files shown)


popey:
Thanks for the PR @abitrolly. I'm in two minds about this. I appreciate extra debug information in the output from quickget. However, in this case, the line immediately after download and check, prints the exact same path. Here's the run with your patch, and note after the `#### 100.0%` line is followed by a line confirming a good sha256sum, while printing out the path.

```
./quickget alpine v3.20
Downloading Alpine Linux v3.20
- URL: https://dl-cdn.alpinelinux.org/alpine/v3.20/releases/x86_64/alpine-virt-3.20.3-x86_64.iso
- PATH: alpine-v3.20/alpine-virt-3.20.3-x86_64.iso
################################################# 100.0%
Checking alpine-v3.20/alpine-virt-3.20.3-x86_64.iso with sha256sum... Good!
Making alpine-v3.20.conf
 - Setting alpine-v3.20.conf executable

To start your Alpine Linux virtual machine run:
    quickemu --vm alpine-v3.20.conf
```

philclifford:
@abitrolly if you omit the v the download fails so there's nothing to check (the release check really could be tighter and should fail before attempting the download). 
@popey at the moment not every installation medium comes with a hash to check so some might not output the path as part of the check.  
I'd be minded to add this little extra output, but I wonder if outputting the apparent absolute rather than the relative path would be better.

quickget
```diff
@@ -1218,6 +1218,7 @@ function web_get() {
     if [[ ${OS} != windows && ${OS} != macos && ${OS} != windows-server ]]; then
         echo "Downloading $(pretty_name "${OS}") ${RELEASE} ${EDITION}"
         echo "- URL: ${URL}"
+        echo "- PATH: ${DIR}/${FILE}"
```
philclifford, line 1221:
```suggestion
        echo "- PATH: ${PWD}/${DIR}/${FILE}"
```
Maybe useful for those with several collections of VMs dotted around the filesystem or on several external drives

lj3954:
> @abitrolly if you omit the v the download fails so there's nothing to check (the release check really could be tighter and should fail before attempting the download). @popey at the moment not every installation medium comes with a hash to check so some might not output the path as part of the check. (see #1545 ) I'd be minded to add this little extra output, but I wonder if outputting the apparent absolute rather than the relative path would be better.

Why do we proceed if the download fails? That's a bug as well.

philclifford (approved):

abitrolly:
@philclifford thanks for merging. I've completely forgot about it.

PR #1565, opened 2025-01-19
fix: Check for PipeWire as well as PulseAudio before falling back to ALSA

### LINKED ISSUE #1564: bug: VMs crashing on Fedora 41 as a result of #1480

**Describe the bug**
On Fedora 41, I'm experiencing an issue on Quickemu 4.9.7 where guest VMs run under Quickemu crash when the guest tries to output audio.

I have narrowed this down to the fix in #1480, which attempts to fix an issue that another user was having when starting VMs on a host without PulseAudio installed.

The logic in the commit merged in #1480 involves using `command -v` to check whether the host has `pacmd` present before falling back to ALSA if the host does not.

I'm not sure this is the most sensible approach here as, if I understand it correctly, Fedora uses PipeWire as the primary audio server now, with compatibility layers for legacy PulseAudio applications. I think a lot of this is done by the `pipewire-pulseaudio` compatibility package which is installed by default on Fedora Workstation.

But what Fedora _doesn't_ have installed by default is any package that provides `pacmd`. So Quickemu 4.9.7 on Fedora 41 (and maybe older) will always default to falling back to ALSA. And I'm not sure the backend plumbing is actually there anymore to handle this, so ultimately this means VMs simply crash whenever they first try and output any audio.

**To Reproduce**
1. Run a Linux VM on Quickemu 4.9.7 on a stock Fedora Workstation 41.
2. Try to output some audio in the VM.
3. Experience the crash.
4. Revert the PR in #1480 and repeat as above.
5. Experience no crash.

**Expected behaviour**
The test in #1480 should probably be rewritten not to make the assumption that any machine without `pacmd` cannot use PulseAudio as an output, to account for distros using PipeWire.

PR to follow with a suggested fix, though I'm open to discussion as it might not be the best overall method of handling things.

alexhaydock:
# Description

- Resolves #1564 

## Type of change

# Checklist:

### DIFF
--- a/quickemu
+++ b/quickemu
@@ -882,7 +882,7 @@ function configure_storage() {
 function configure_display() {
     # Determine which audio driver use between Pulseaudio or ALSA
     local AUDIO_DRIVER="pa"
-    if ! command -v pacmd >/dev/null 2>&1 ; then
+    if ! command -v pacmd >/dev/null 2>&1 && ! command -v pipewire-pulse >/dev/null 2>&1; then
         AUDIO_DRIVER="alsa"
     fi
 
--- a/quickemu
+++ b/quickemu
@@ -882,7 +882,12 @@ function configure_storage() {
 function configure_display() {
     # Determine which audio driver use between Pulseaudio or ALSA
     local AUDIO_DRIVER="pa"
-    if ! command -v pacmd >/dev/null 2>&1 && ! command -v pipewire-pulse >/dev/null 2>&1; then
+    if pidof pipewire >/dev/null 2>&1; then
+        # QEMU's pipewire audio backend was added in version 8.1
+        if [ "${QEMU_VER_SHORT}" -ge 81 ]; then
+            AUDIO_DRIVER="pipewire"
+        fi
+    elif ! pidof pulseaudio >/dev/null 2>&1; then
         AUDIO_DRIVER="alsa"
     fi
 


szorfein:
This will not work for pure Alsa system. Pipewire is always installed as dependencies for another app even if not used.

alexhaydock:
A good point, though I suppose that would have similar downsides when running Quickemu on non-systemd distros like Alpine or Void.

I guess this check might need to get more complex and layered.

szorfein:
if work for me with if u can test

    if ! command -v pacmd >/dev/null 2>&1 && ! command -v pipewire-pulse >/dev/null 2>&1; then

alexhaydock:
> if work for me if testing with `pipewire-pulse` if u can test
> 
> ```
> if ! command -v pacmd >/dev/null 2>&1 && ! command -v pipewire-pulse >/dev/null 2>&1; then
> ```

Thanks! Have updated the PR to look for `pipewire-pulse` instead of just `pipewire`.

I think that's definitely more robust and I can confirm it works to solve the original bug still on Fedora 41. Should also cope with situations where the system is pure ALSA still, like you mentioned before.

lj3954:
Could we not just check whether one of these servers is running? 
```
if !pidof pipewire 2>&1 /dev/null && ! pidof pulseaudio 2>&1 /dev/null; then
```

quickemu
```diff
@@ -882,7 +882,7 @@ function configure_storage() {
 function configure_display() {
     # Determine which audio driver use between Pulseaudio or ALSA
     local AUDIO_DRIVER="pa"
-    if ! command -v pacmd >/dev/null 2>&1 ; then
+    if ! command -v pacmd >/dev/null 2>&1 && ! command -v pipewire-pulse >/dev/null 2>&1; then
         AUDIO_DRIVER="alsa"
     fi
```
lj3954, line 887:
```suggestion
    if pidof pipewire >/dev/null 2>&1; then
        # QEMU's pipewire audio backend was added in version 8.1
        if [ "${QEMU_VER_SHORT}" -ge 81 ]; then
            AUDIO_DRIVER="pipewire"
        fi
    elif ! pidof pulseaudio >/dev/null 2>&1; then
        AUDIO_DRIVER="alsa"
    fi
```

lj3954 (commented):
QEMU 8.1 added a pipewire audio backend. We should use it if possible. This should be a better solution in general.

quickemu
szorfein: good idea to test pipewire and pulseaudio with `pidof` and this include pipewire to the project :)

alexhaydock:
> QEMU 8.1 added a pipewire audio backend. We should use it if possible. This should be a better solution in general.

This looks good to me. Makes sense to actually use PipeWire natively if available in both QEMU and on the host. My testing on Fedora 41 suggests this works perfectly.

Having accepted your suggestion directly within the GitHub UI, there are now two commits with different solutions to the original issue so when merging it will make sense to squash them.

lj3954 (approved):

PR #1575, opened 2025-01-29
fix: Allow non-v24 arcolinux releases

### LINKED ISSUE #1574: bug: Quickget does not display new arcolinux v25 release
Would be nice if this is changed to show the last 3 or 5 releases. 

Also, what can I do till this gets solved ?

lj3954:
# Description

Quickget was modified to only allow arcolinux versions with the major release being 24 and minor being 06 or above, due to a change in editions taking place at that point. This is no longer necessary, as the latest 5 releases are well past this change. Version 25 has been released and isn't matched by the current regex.

- Resolves #1574

## Type of change

# Checklist:

### DIFF
--- a/quickget
+++ b/quickget
@@ -545,9 +545,7 @@ function releases_archlinux() {
 
 function releases_arcolinux() {
     #shellcheck disable=SC2046,SC2005
-    # breaking change in v24.05
-    # v24.05.1 is the first release with the new naming scheme and too complex to parse old and new so just show the new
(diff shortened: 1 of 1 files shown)


philclifford (approved):

arcolinuxz:
Wrong url guys

Trust me - from the developer himself
https://ant.seedhost.eu/arcolinux/iso/

this server sees the ISOs first
rest a day or more later

erikdubois:
Working on this version on Arch 

image

We can not download the ISO.

It says it is finished but it is not.

PR #1579, opened 2025-02-01
fix: Secure Boot does not work for Linux guests

### LINKED ISSUE #413: feat: support secure boot with Microsoft keys
Currently, Quickemu supports enabling Secure Boot with `secureboot=on`, and this causes a Secure Boot compatible UEFI image to be loaded.

Example below is what gets picked on Fedora:
https://github.com/quickemu-project/quickemu/blob/3940f7b4521ee840c6811dde120734df97dc0e07/quickemu#L390

But this image doesn't contain the actual default Microsoft Platform Keys for Secure Boot - instead, they're loaded into the NVRAM, which gets stored in `OVMF_VARS.fd`. Using Fedora as an example, this file gets sourced from here:
https://github.com/quickemu-project/quickemu/blob/3940f7b4521ee840c6811dde120734df97dc0e07/quickemu#L391

But that NVRAM image _isn't_ the one with the keys enrolled in it. That image is stored here:
```
/usr/share/edk2/ovmf/OVMF_VARS.secboot.fd
```

So this means that, functionally, using `secureboot=on` doesn't actually turn Secure Boot on in the way a user might expect, since Secure Boot **is** available in the UEFI but is disabled (because it doesn't have any keys to validate against).

If I change line 391 (linked above) to the following:
```
          efi_vars "/usr/share/edk2/ovmf/OVMF_VARS.secboot.fd" "${EFI_VARS}"
```
this works fine. The UEFI has the Microsoft Platform Key(s) enrolled, and Secure Boot is enabled by default.

Is it worth me submitting a PR to add support for this?

**WARNING**
There's a potential caveat to be aware of though before we pull in such a PR - which is that if you're using Quickemu to run Windows and Secure Boot is properly enabled, the [Windows Driver Signing Policy requirements](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/kernel-mode-code-signing-policy--windows-vista-and-later-#signing-requirements-by-version) demand that you have WHQL-signed drivers blessed by Microsoft's keys in order for them to be allowed to load.

This might cause some confusion or annoyance if we pull this in without any other safeguards, because the common strategy of downloading the `virtio-win.iso` from the FedoraPeople site won't work. The drivers for the VirtIO components are signed by Red Hat's keys, but they're not WHQL-signed by Microsoft.

But Red Hat _does_ actually make these available as part of the `virtio-win` package. And they work completely fine. I just think you have to be a paying RHEL subscriber to be able to get those ones.

**TESTED AND WORKING**
Here's an example of fully-working Secure Boot on Windows 10 inside Quickemu on Fedora 35, with only the one-line change above (and the VirtIO drivers from the `virtio-win` package from a paid RHEL subscription):
image

### LINKED ISSUE #1578: bug: Secure Boot does not work for Linux guests

**Describe the bug**
When booting Linux guests, enabling Secure Boot with `secureboot="on"` appears to stop the VM from booting.

I have tested this on Linux hosts, but not on any other host platforms so mileage may vary there.

This is partially related to my issue #413.

PR to follow with some suggested steps to solve this, though it may need some work. The PR should also partially fix #413.

**To Reproduce**
Run a Quickemu guest with the following `.conf` (example uses Fedora but it works for all Linux distros):
```ini
#!/usr/bin/quickemu --vm
guest_os="linux"
bios_type="efi"
secureboot="on"
disk_img="test/disk.qcow2"
iso="test/Fedora-Workstation-Live-x86_64-41-1.4.iso"
```

Experience the fact that the VM does not provide any display output:

Image

**Expected behaviour**
VM should boot any Linux guest that has Secure Boot support (Debian, Fedora, Ubuntu, etc) without issues, and report that Secure Boot is Enabled when running `mokutil --sb-state` (which indicates that Secure Boot is not only working, but that the Microsoft Platform Keys are enabled as per #413).

Image

**System information**
Running latest Quickemu from `master` on Fedora 41 Workstation.

**Additional context**
PR to follow.

alexhaydock:
# Description

This change contains two fixes which ought to take the functionality of Secure Booting Linux guests in Quickemu from non-functional to a working state.

### SMM fix
Commit b9735800f3b2b5d66aec6e984db9d43cb6e9c4c3 should stand alone and resolve #1578. This enables SMM for Linux guests on Linux hosts when Secure Boot has been enabled.

### Microsoft Platform Key fix (partial)
Commit 8add3185e075c466a0d0692d5c87f1d5ad671362 is an example of an initial commit which will probably need wider expansion and partially resolves #413. This configures Quickemu to prefer the `*.secboot.*` version of the `OVMF_VARS` `.fd` file, which is the one which has the Microsoft Platform Keys preloaded into it.

I'm honestly less sure about this second fix as although it makes sense to me (the `*.secboot.*` file is needed to have _full_ Secure Boot working, rather than just Secure Boot stuck in Platform Setup Mode), I note that Quickemu also supports Windows 11 which makes Secure Boot mandatory, but I couldn't work out how that support was being provided.

I'll do some digging, but it's possible that the Secure Boot config for Windows 11 also hasn't been using the Microsoft Platform Keys up until now, meaning it is effectively doing nothing in Quickemu's current state.

This commit will also need expanding. Currently not every distribution ships the Microsoft PK enrolled VARS files, and the ones that do seem to put them in different places. I'm happy to go hunting around the common host distros that people might be using for Quickemu if this commit is reviewed as sensible and expand the search list for these files with pre-enrolled keys, but for now this fix only applies to Fedora as a host distro (or anything else that uses the exact same path on disk as Fedora does for these files).

- Closes #413 
- Resolves #1578

## Type of change

# Checklist:

  - Presented as a draft for now - more testing needed once code is reviewed as generally sensible

### DIFF
--- a/quickemu
+++ b/quickemu
@@ -673,7 +673,7 @@ function configure_bios() {
         if [ -n "${EFI_CODE}" ] || [ ! -e "${EFI_CODE}" ]; then
             case ${secureboot} in
                 on) # shellcheck disable=SC2054,SC2140
-                    ovmfs=("${SHARE_PATH}/OVMF/OVMF_CODE_4M.secboot.fd","${SHARE_PATH}/OVMF/OVMF_VARS_4M.fd" \
+                    ovmfs=("${SHARE_PATH}/OVMF/OVMF_CODE_4M.secboot.fd","${SHARE_PATH}/OVMF/OVMF_VARS_4M.ms.fd" \
                         "${SHARE_PATH}/edk2/ovmf/OVMF_CODE.secboot.fd","${SHARE_PATH}/edk2/ovmf/OVMF_VARS.secboot.fd" \
                         "${SHARE_PATH}/OVMF/x64/OVMF_CODE.secboot.fd","${SHARE_PATH}/OVMF/x64/OVMF_VARS.fd" \
                         "${SHARE_PATH}/edk2-ovmf/OVMF_CODE.secboot.fd","${SHARE_PATH}/edk2-ovmf/OVMF_VARS.fd" \
--- a/quickemu
+++ b/quickemu
@@ -395,6 +395,16 @@ function configure_cpu() {
         fi
     fi
 
+    # SMM is also required for Linux guests when Secure Boot is enabled
+    if [ "${secureboot}" == "on" ]; then
+        if [ "${guest_os}" == "linux" ]; then
+        # SMM is not available on QEMU for macOS via Homebrew
+            if [ "${OS_KERNEL}" == "Linux" ]; then
+                SMM="on"
+            fi
+        fi
+    fi
+
     case ${guest_os} in
         batocera|freedos|haiku|solaris) MACHINE_TYPE="pc";;
         kolibrios|reactos)
--- a/quickemu
+++ b/quickemu
@@ -674,7 +674,7 @@ function configure_bios() {
             case ${secureboot} in
                 on) # shellcheck disable=SC2054,SC2140
                     ovmfs=("${SHARE_PATH}/OVMF/OVMF_CODE_4M.secboot.fd","${SHARE_PATH}/OVMF/OVMF_VARS_4M.fd" \
-                        "${SHARE_PATH}/edk2/ovmf/OVMF_CODE.secboot.fd","${SHARE_PATH}/edk2/ovmf/OVMF_VARS.fd" \
+                        "${SHARE_PATH}/edk2/ovmf/OVMF_CODE.secboot.fd","${SHARE_PATH}/edk2/ovmf/OVMF_VARS.secboot.fd" \
                         "${SHARE_PATH}/OVMF/x64/OVMF_CODE.secboot.fd","${SHARE_PATH}/OVMF/x64/OVMF_VARS.fd" \
                         "${SHARE_PATH}/edk2-ovmf/OVMF_CODE.secboot.fd","${SHARE_PATH}/edk2-ovmf/OVMF_VARS.fd" \
                         "${SHARE_PATH}/qemu/ovmf-x86_64-smm-ms-code.bin","${SHARE_PATH}/qemu/ovmf-x86_64-smm-ms-vars.bin" \


quickemu
```diff
@@ -665,6 +675,7 @@ function configure_bios() {
                 on) # shellcheck disable=SC2054,SC2140
                     ovmfs=("${SHARE_PATH}/OVMF/OVMF_CODE_4M.secboot.fd","${SHARE_PATH}/OVMF/OVMF_VARS_4M.fd" \
                         "${SHARE_PATH}/edk2/ovmf/OVMF_CODE.secboot.fd","${SHARE_PATH}/edk2/ovmf/OVMF_VARS.fd" \
+                        "${SHARE_PATH}/edk2/ovmf/OVMF_CODE.secboot.fd","${SHARE_PATH}/edk2/ovmf/OVMF_VARS.secboot.fd" \
```
lj3954, line 678: Why is this here? It won't do anything. Quickemu searches the OVMF code paths (even indices in the array) and selects the vars (code index+1) based on it. You've added an identical OVMF code path to a prior entry, which will never be matched. Also, which distro has this OVMF_VARS.secboot.fd file? I can't find it in Ubuntu or Arch's edk packages.

lj3954: Oh, I see. It's supposed to resolve a different issue. It won't work, though. Please address the above concerns though, we can't have quickemu pointing to nonexistent files on any distro.

alexhaydock: > Why is this here? It won't do anything.

This includes the Microsoft PK keys. Without using the file pre-seeded with these keys, Secure Boot isn't actually doing anything when enabled. It just sits in Platform Setup mode, doing no validation.

> Quickemu searches the OVMF code paths (even indices in the array) and selects the vars (code index+1) based on it. You've added an identical OVMF code path to a prior entry, which will never be matched.

I could be mistaken but, based on my testing, it seemed like this list gets evaluated in reverse-order so this ought to take precedence over the entry in the line above it. I think this is the desirable behaviour.

> Also, which distro has this OVMF_VARS.secboot.fd file?

This path, as well as the one in the line above it come from `edk2-ovmf` shipped by Fedora (and EL8, EL9, EL10): https://src.fedoraproject.org/rpms/edk2/blob/f41/f/edk2.spec#_664

> I can't find it in Ubuntu or Arch's edk packages.

From what I can tell, Arch doesn't ship the MS keys. Ubuntu seems to use `/usr/share/OVMF/OVMF_VARS_4M.ms.fd` as the path to the VARS pre-seeded with MS keys.

> we can't have quickemu pointing to nonexistent files on any distro.

Why not? Most of these files don't exist on most of the distros that a user will run Quickemu on. Surely that's the whole point of iterating through the list?

alexhaydock: Looking closer, I think I see what you mean.

The `if [ -e "${1}" ]` loop checks only the existence of the CODE file, making the existence of the VARS file irrelevant?

I had initially read this as if it checked **both** the CODE file and VARS file.

This fix does actually produce the desired behaviour but presumably what is happening is that because there's no early abort/return on the `for` loop, this code will always match the last path to a CODE file in the list that exists, in a situation where multiple exist. This explains why I thought it was being evaluated in reverse-order.

I added this as a new line rather than updating the line in-place because although my view is that it will _always_ be preferable to use the VARS file with pre-seeded keys if it exists, I didn't want to cause a regression by updating the path reference in-place, in case there were distros that shipped `/usr/share/edk2/ovmf/OVMF_VARS.fd` but not the equivalent `/usr/share/edk2/ovmf/OVMF_VARS.secboot.fd`.

Based on some searching though, it appears this only applies to one distro (Amazon Linux 2, for some reason), and users are unlikely to be running Quickemu on that:
* https://pkgs.org/search/?q=%2Fusr%2Fshare%2Fedk2%2Fovmf%2FOVMF_VARS.fd
* https://pkgs.org/search/?q=%2Fusr%2Fshare%2Fedk2%2Fovmf%2FOVMF_VARS.secboot.fd

alexhaydock: FWIW, updating the equivalent path in-place for Debian/Ubuntu hosts would appear to be safe, as there are no major distros I can find that ship the non-PK-seeded file but don't ship the PK-seeded one.

See:
* https://pkgs.org/search/?q=%2Fusr%2Fshare%2FOVMF%2FOVMF_VARS_4M.fd
* https://pkgs.org/search/?q=%2Fusr%2Fshare%2FOVMF%2FOVMF_VARS_4M.ms.fd

alexhaydock: Based on some searching:

### Distros which ship the MS PK keys in a separate VARS file:
* Debian - `/usr/share/OVMF/OVMF_VARS_4M.ms.fd`
* Ubuntu - `/usr/share/OVMF/OVMF_VARS_4M.ms.fd`
* Fedora - `/usr/share/edk2/ovmf/OVMF_VARS.secboot.fd`
* CentOS / RHEL / etc - `/usr/share/edk2/ovmf/OVMF_VARS.secboot.fd`

### Distros which do not seem to ship the MS keys at all
* Alpine - `/usr/share/OVMF/OVMF_VARS.fd`
* Arch - `/usr/share/edk2/x64/OVMF_VARS.4m.fd`
* Void - `/usr/share/edk2/x64/OVMF_VARS.4m.fd`

### Distros which only ship one VARS file (which always has the MS keys in it)
* NixOS
  * Based on what I can tell from [the nixpkgs entry for ovmf](https://github.com/NixOS/nixpkgs/blob/666e1b3f09c267afd66addebe80fb05a5ef2b554/pkgs/applications/virtualization/OVMF/default.nix#L207)

alexhaydock: Updated to use a replace-in-place method based on the above validation.

philclifford (approved):

Issue #1648, opened 2025-05-12
ERROR! QEMU 6.0.0 or newer is required, detected 10.0.0.


Cris70:
**I confirm this bug has not already been reported**
- [X] I have searched the issues and this bug has not been reported previously

**Describe the bug**
When I run quickemu (with any option, it doesn't matter) I get the error in the subject

**To Reproduce**
Steps to reproduce the behaviour:
1. Run `quickemu` with e.g. arguments '--version' (but you can use any argument)
2. See error

**Expected behaviour**
Quickemu should correctly compare version numbers and run without issues

**Quickemu output**
Run `quickemu`  and include the output of the failure below:

<details>
  <summary>Quickemu output</summary>

  ```text
   ERROR! QEMU 6.0.0 or newer is required, detected 10.0.0.
  ```
</details>

**System information**
Run `quickreport` and include the output here; if you can't run `quickreport`,
please provide the output of the following:

<details>
  <summary>Quickreport output</summary>

  ```text
----------------------------------
        Quickemu ERROR! QEMU 6.0.0 or newer is required, detected 10.0.0.
----------------------------------
Distro:	openSUSE Tumbleweed
Kernel:	Linux 6.14.5-1-default x86_64
Memory:	33G
GPU:
 - Advanced Micro Devices, Inc. [AMD/ATI] Lexa PRO [Radeon 540/540X/550/550X / RX 540X/550/550X] (rev c7)
----------------------------------
            curl 8.13.0
----------------------------------
Libraries: libcurl/8.13.0 OpenSSL/3.5.0 zlib/1.3.1 brotli/1.1.0 zstd/1.5.7 libidn2/2.3.8 libpsl/0.21.5 libssh/0.11.1/openssl/zlib nghttp2/1.65.0 OpenLDAP/2.6.8
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp ws wss
Features:  alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd
----------------------------------
            QEMU 10.0.0
----------------------------------
Available CPUs:
  486                   (alias configured by machine type)
  486-v1                
  Broadwell             (alias configured by machine type)
[...]

PR #1684, opened 2025-07-05
fix: warning about microsoft blocking your IP now specifies it may be a range block if using IPv6

BPplays:
# Description

when i was trying to download windows 11 my download got blocked and it said:
*WARNING! Microsoft blocked the automated download request based on your IP address.* but when i changed my IPv6 address to another one in my prefix delegation it still didn't work suggesting they probably block a /64 at a time

so i changed the wording to say *WARNING! Microsoft blocked the automated download request based on your IP address or IP range.*

btw www.microsoft.com does have a AAAA IPv6 DNS record

## Type of change

# Checklist:

- [ ] I have performed a self-review of my code

### DIFF
--- a/quickget
+++ b/quickget
@@ -3287,7 +3287,7 @@ function download_windows_workstation() {
     fi
 
     if echo "$iso_download_link_json" | grep -q "Sentinel marked this request as rejected."; then
-        echo " - WARNING! Microsoft blocked the automated download request based on your IP address."
+        echo " - WARNING! Microsoft blocked the automated download request based on your IP address or IP range."
         failed=1
     fi
 


alexhaydock:
@BPplays Are you sure the download is actually coming from a server with an AAAA record?

One thing I noticed is that MS blocks me from downloading when I'm using a public NAT64 proxy (so where I've got a shared IPv4 exit) but this block goes away when I switch to using my own, which adds v4 from my own ranges at the edge of my network.

So I suspect the actual download server might be v4 only in this case.

BPplays:
@alexhaydock 
i just tested with `curl -6` in https://github.com/quickemu-project/quickemu/blob/9f90d46ea195d62a8b84c6dd45ea5b1443bb3e46/quickget#L3279 and https://github.com/quickemu-project/quickemu/blob/9f90d46ea195d62a8b84c6dd45ea5b1443bb3e46/quickget#L1266
and it seems to download fine

also it would be kinda strange to block IPv6 on the download because you'd need to rely on anything using the api falling back to v4 since the domain it's going to has a AAAA record

BPplays:
@alexhaydock 
i just tested with `curl -6` in https://github.com/quickemu-project/quickemu/blob/9f90d46ea195d62a8b84c6dd45ea5b1443bb3e46/quickget#L3279 and https://github.com/quickemu-project/quickemu/blob/9f90d46ea195d62a8b84c6dd45ea5b1443bb3e46/quickget#L1266
and it seems to download fine
i am using dns64 too but it shouldn't generate a dns64 record if the domain has a AAAA already

also it would be kinda strange to block IPv6 on the download because you'd need to rely on anything using the api falling back to v4 since the domain it's going to has a AAAA record

alexhaydock:
Looks like the actual download domain is `software.download.prss.microsoft.com` but that does also have an AAAA record. Interesting. I wonder why my NAT64 gateway ended up being used in the past.

Edit: Ah it's this request that gets blocked for me I think
https://github.com/quickemu-project/quickemu/blob/9f90d46ea195d62a8b84c6dd45ea5b1443bb3e46/quickget#L3257

`vlscppe.microsoft.com` doesn't have an AAAA record so this API query goes via NAT64.

BPplays:
@alexhaydock ah alright i assumed that since it checks `iso_download_link_json` to display the warning that the call to curl www.microsoft.com is what would trigger it

alexhaydock:
@BPplays Out of interest - is your NAT64 gateway a public/shared one? This might still end up being a useful change, or something useful to call out in the docs - it's just an interesting theory I have.

BPplays:
@alexhaydock my nat64 gateway isn't public.
the reason i noticed this in the first place was i think because of #1648 and quickgui not showing the error (or maybe i didn't notice) so i re-downloaded windows 11 a couple of times before i realized why it wouldn't start when i used quickemu

PR #1735, opened 2025-09-18
fix: restore rockylinux dynamic releases

philclifford:
# Description

The website was changed and the old parse was not cutting it

- [x] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [x] I have performed a self-review of my code
- [x] I have tested my code in common scenarios and confirmed there are no regressions

PR #1738, opened 2025-09-21
Feat 1722

### LINKED ISSUE #1722: feat: toggles to reduce attack surface
- [ x ] I have searched the issues and this feature has not previously been requested

Quickemu as a project aims to "do the right thing" when setting up VMs so they "just work".

This is definitely a valuable goal, but it's such a good software project that it's useful for many use-cases, including situations where a VM is being used to test less-trusted software inside a guest.

In these situations it would be nice to be able to reduce some of the automatically-enabled features of guest VMs with simple boolean toggles in the config file.

From what I can tell, the following features are added automatically to the QEMU commandline executed behind the scenes and are not easily toggleable to be "off":
* USB/SPICE redirection (regardless of whether USB passthrough is actually being used)
* USB CCID / Smartcard passthrough (regardless of whether the user actually needs one)
* Folder sharing between host and guest with 9p
* SSH forwarding

In my view, it ought to be reasonably easy to implement a boolean toggle for these options in the config file. Something along the lines of:
```ini
usb_redir="false"
usb_ccid="false"
folder_sharing="false"
ssh_fwd="false"
```

To maintain the "it just works" approach of Quickemu I think the best approach would be to retain the default-on mode of these features, and just document the boolean off toggles for them in the Advanced Features section of the Wiki.

**Describe alternatives you've considered**
N/A

This approach seems sensible and is not a breaking change as Quickemu will continue to work exactly as before for users who do not enable the above options in their config files.

**Additional context**
N/A

philclifford:
# Description

adds '--paranoid' option (also triggered by setting QUICKEMU_PARANOID in the environment ). Also individual config file toggles are available for :
 - usb_redir="false"
 - usb_ccid="false"
 - folder_sharing="false"
 - ssh_fwd="false"

These default to "true"  to retain the current behaviour.

- Closes #1722
- Fixes #1735 (so that documentation changes are consistent)

## Type of change

- [x] Documentation (updates the documentation)

# Checklist:

- [x] I have made corresponding changes to the documentation


alexhaydock:
Looking a bit deeper into this one, it's mostly working as expected, but I think it might be adding some stray config for the file sharing into the bash start-scripts.

I can see the paranoid skip has been added to some of the file sharing setup here:
https://github.com/philclifford/quickemu/blob/35293ec7b8cce859df087d48e40c753c383a7aad/quickemu#L1214-L1219

But I think this section here might need it too, otherwise some of the options still end up in the `.sh` script:
https://github.com/philclifford/quickemu/blob/35293ec7b8cce859df087d48e40c753c383a7aad/quickemu#L1485-L1489

philclifford:
Good catch - thanks.  I'll try and tweak that bit too when I'm awake.

philclifford:
OK @alexhaydock I think I've done the necessary, could you take another look please.

flexiondotorg:
Thank you for the work you've put into this, Phil - I really appreciate the effort and care you've taken with the implementation.

After some reflection, I've decided not to merge this. I've explained my reasoning in more detail on the original issue (#1722), but the short version is that I'm concerned about scope creep and want to keep quickemu focused on "just works" rather than adding configuration toggles.

I hope you understand, and thank you again for contributing.

Issue #615, opened 2023-01-20
Support for CPU pinning


UtkarshVerma:
I wish to be able to pin some cores of my processor to the QEMU process. Would it be possible to get that done with `quickemu`?

philclifford:
You can add additional qemu arguments via a parameter in the VM.conf. So you can already achieve that I believe, but you should do a bit of research, consider the goals you are aiming for, be aware of the impact of your CPU architecture and numa arrangements... You might also need to tweak settings on your host if that is necessary to, say, reserve some cores/threads for exclusive VM use (as opposed to forcing the VM virtual cores/threads to specific physical cores/threads (/chiplets/CPUs ...)

akarasulu:
@philclifford Which parameters should we use specifically within the VM.conf to pass through CPU pinning information. 

Performance is terrible **WITHOUT** CPU pinning especially in macOS and Windows. This should certainly be addressed. In fact CPU pinning should be enabled out of the box or you're looking at inordinate amounts of time in both OS installations.

There's a lot of greatness in quickemu however this problem draws away from that. Should be high priority to solve for baseline usability.

9SMTM6:
I did look around a decent bit. AFAICT from `man quickemu` and `man quickemu_conf` (also as provided in this very repo) you can not provide generic arguments to `qemu-system-XXX`, which is of course what you'd need to do.

You cant even provide the number of threads, which is easy enough with raw qemu.

How to do what is commonly referred to as cpu-pinning with the raw commandline isnt very well documented in secondary documentation (articles etc, they mostly suggest using `libvirt` with qemu/kvm, sometimes additionally wrapped in something like virt-manager). But as far as I can tell we should be able to do it or at least something very similar with `qemu-system-x86_64 -numa` (one can read the fitting section in `man qemu-system-...` as reference). In the end libvirt uses qemu/kvm in that config so it should be possible.

But as it stands, as far as I can tell, quickemu does NOT support this even indirectly by providing a way to pass through parameters.

The best way we can currently use quickemu with it, as far as I can tell, is to make it generate a windows machine, calling it once with quickemu, and then afterwards copy and manipulate the generated  `<VM>.sh` instead.

[...]

PR #1771, opened 2025-11-20
feat: add support for CPU pinning

ivatolm:
# Description

Added support for pinning host CPU cores to logical guest cores.

- Resolves (or improves upon) #615 

## Implementation
I've added a new parameter (`--cpu-pinning`) that specifies whether Quickemu should pin guest cores to any particular host cores.  Parameter accepts 1 positional argument - comma-separated list of host cores. Number of host cores should match number of logical cores Quickemu decided to use for the VM.

## Why only logical cores and not all guest cores?
To me it seems more logical to preserve actual hardware core count that would've been used in the case of bare metal. For example, if hyperthreading is enabled and we use 2 cores, we get 4 vCPUs, but only 2 actual physical cores would be doing the work.

**(I am open to discussion about whether this is a right approach or not)**

Example: 
`quickemu --vm alpine-v3.22.conf --cpu-pinning 5,6`
(Quickemu decided to use 2 cores for my VM, so I've specified 2 cores)

## Type of change

# Checklist:

- [ ] I have made corresponding changes to the documentation (decided to agree on design first, will add later)

I've tested this code in multiple scenarios, but if you have any suggestions for edge cases that might break with this change, I’d be happy to test those as well!

I'm non-native speaker, please let me know if anything needs further clarification!

### DIFF
--- a/quickemu
+++ b/quickemu
@@ -1156,11 +1156,41 @@ function configure_tpm() {
     fi
 }
 
+function configure_cpu_pinning() {
+    if [ -z "${CORE_MAPPING}" ]; then
+        return
+    fi
+
+    GUEST_CPUS=""
+    idx=0
+    for tid_dir in /proc/"${VM_PID}"/task/*; do
+        tid=$(basename "$tid_dir")
+        name=$(cat "$tid_dir/comm")
+
+        if [[ "$name" == CPU* ]]; then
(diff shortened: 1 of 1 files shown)


flexiondotorg:
@cubic-dev-ai Review this

PR #1795, opened 2026-01-14
fix(nix): glxinfo has been renamed mesa-demos

### LINKED ISSUE #1763: bug: glxinfo has been renamed to mesa-demos in package.nix

**Describe the bug**
Using flake.nix is broken, for official unstable nixpkgs. This is currently not an issue if user is using flake.nix without overriding to their local (latest unstable) nixpkgs. But eventually this will be an issue also here - when someone updates flake.lock inputs with latest one.

**To Reproduce**
1. Use flake.nix in your own flake:

```nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
    quickemu = {
      url = "github:quickemu-project/quickemu";
      inputs.nixpkgs.follows = "nixpkgs";  # using your own unstable nixpkgs
    };
  };
}
```

2. See error:

```
error: 'glxinfo' has been renamed to/replaced by 'mesa-demos'
```

**Expected behaviour**
To not get the error.

**Additional info**
The fix is just to rename the offending package (and to update nixpkgs in flake.lock):
- here https://github.com/quickemu-project/quickemu/blob/master/package.nix#L10
- and here https://github.com/quickemu-project/quickemu/blob/master/package.nix#L54

But since I believe glxinfo executable is just a test program which is usually used only for checking how many FPS do you get... So is this even required, could we get rid of this dependency?

supermarin:
The nix package was renamed, but binary stays the same

# Description

Please include a summary of the changes along with any relevant motivation and context.

- Closes #1763 

## Type of change

- [x] Packaging (updates the packaging)

# Checklist:

### DIFF
--- a/devshell.nix
+++ b/devshell.nix
@@ -26,7 +26,7 @@ mkShell {
     OVMF
     OVMFFull
   ] ++ lib.optionals stdenv.isLinux [
-    glxinfo
+    mesa-demos
     usbutils
     xdg-user-dirs
   ]);
--- a/package.nix
+++ b/package.nix
@@ -7,10 +7,10 @@
 , cdrtools
 , curl
 , gawk
-, glxinfo
 , gnugrep
 , gnused
 , jq
+, mesa-demos
 , pciutils
 , procps
 , python3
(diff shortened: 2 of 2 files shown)


flexiondotorg:
@cubic-dev-ai Review this

PR #1824, opened 2026-01-25
perf(quickemu): tune TCG translation cache and enable multithreaded TCG

flexiondotorg:
- Add TCG-specific runtime optimisations in vm_boot for cross-architecture VMs
- Detect host RAM and set tb-size to 512 for hosts with >=16GB, otherwise 256
- Append -accel tcg,tb-size=${TCG_TB_SIZE},thread=multi to QEMU args
- Improve TCG translation cache behaviour and SMP performance for TCG guests

## Type of change

# Checklist:


flexiondotorg:
@cubic-dev-ai

PR #1847, opened 2026-01-26
feat(quickemu): enable SPICE display reconnection for running VMs

### LINKED ISSUE #1370: feat: Connect to a detached running vm
- [x] I have searched the issues and this feature has not previously been requested

It is difficult to connect to a running vm (started by qemu), when the window is terminated in cases like a compositor crash. In that case, the vm is still running. Trying to start the vm reports:
```bash
virtual-machine
 - Process:  Already running ./virtual-machine.conf as virtual-machine (<a pid of qemu-system-x86_64>)
```

It will be helpful to connect to a detached virtual machine using a flag or just by running the connect command.

flexiondotorg:
Fixes #1370

## Type of change

# Checklist:

### DIFF
--- a/quickemu
+++ b/quickemu
@@ -165,10 +165,12 @@ function kill_vm() {
     if [ -z "${VM_PID}" ]; then
         echo " - ${VMNAME} is not running."
         rm -f "${VMDIR}/${VMNAME}.pid"
+        rm -f "${VMDIR}/${VMNAME}.spice"
     elif [ -n "${VM_PID}" ]; then
         if kill -9 "${VM_PID}" > /dev/null 2>&1; then
             echo " - ${VMNAME} (${VM_PID}) killed."
             rm -f "${VMDIR}/${VMNAME}.pid"
+            rm -f "${VMDIR}/${VMNAME}.spice"
         else
             echo " - ${VMNAME} (${VM_PID}) was not killed."
         fi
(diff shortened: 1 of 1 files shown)


flexiondotorg:
Thanks for the review, but I believe the implementation is correct. Let me explain the execution flow:

## Scenario 1: Starting a NEW VM

1. `VM_PID` is empty (no running process)
2. Code takes the `if [ -z "${VM_PID}" ]` branch (line 2789)
3. `vm_boot` is called → `configure_ports` runs
4. `configure_ports` deletes any stale `.spice` file (line 1414) — **correct behaviour**
5. If SPICE is enabled, a new `.spice` file is created with the new port (line 1478)

## Scenario 2: Reconnecting to an ALREADY RUNNING VM

1. Lines 2734-2742 check the `.pid` file and validate the process is running via `kill -0`
2. `VM_PID` is set to the running process ID
3. Code takes the `else` branch (line 2800) — **`vm_boot` is never called**
4. Since `configure_ports` is only called from within `vm_boot`, it **never runs** for reconnection
5. The `.spice` file remains intact
6. Lines 2805-2809 read the `.spice` file and set `display="spice"`
7. `start_viewer` launches the SPICE viewer

The key insight: `configure_ports` is only invoked when booting a new VM. The reconnection path bypasses `vm_boot` entirely, so the `.spice` file is preserved and correctly used for reconnection.

PR #1860, opened 2026-01-28
feat: add support for passing extra arguments to the SPICE viewer

lvillani:
# Description

This PR adds support for passing additional arguments to the SPICE viewer. The feature was added mainly to support passing `--spice-usbredir-redirect-on-connect` in order to support automatic USB device redirection. For example (in a configuration file):

```bash
viewer_extra_args="--spice-usbredir-redirect-on-connect=-1,0x1234,0x5678,-1,1"
```

Where `0x1234` and `0x5678` are example USB VID and PID, respectively.

The string format is described at <https://www.spice-space.org/usbredir.html#filter-string-format>.

This PR also contains a minor style commit (removing trailing whitespace) and another commit clarifying that "viewer" actually refers to the SPICE viewer.

## Type of change

- [x] Documentation (updates the documentation)

# Checklist:

- [x] I have made corresponding changes to the documentation

### DIFF
--- a/docs/quickemu.1.md
+++ b/docs/quickemu.1.md
@@ -259,7 +259,7 @@ These should be handled by dependencies in Trixie and later. For earlier
 versions (and their derivatives):
 
 ``` shell
-sudo apt-get install bash coreutils curl genisoimage grep jq mesa-utils ovmf pciutils procps python3 qemu sed socat spice-client-gtk swtpm-tools unzip usbutils util-linux xdg-user-dirs xrandr zsync 
+sudo apt-get install bash coreutils curl genisoimage grep jq mesa-utils ovmf pciutils procps python3 qemu sed socat spice-client-gtk swtpm-tools unzip usbutils util-linux xdg-user-dirs xrandr zsync
 ```
 
 #### Install requirements on Fedora hosts
--- a/docs/quickemu.1.md
+++ b/docs/quickemu.1.md
@@ -75,7 +75,7 @@ You can also pass optional parameters
 :   Do not commit any changes to disk/snapshot.
 
 **--viewer \<viewer\>**
-:   Choose an alternative viewer. @Options: 'spicy' (default),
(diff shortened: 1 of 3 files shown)


lvillani:
I updated commit messages to follow the current style a bit more closely (namely, I added the `quickemu` scope).

I'm a bit unsure about the remark from cubic since the feature is implemented nearly the same was as `--extra-args`.

flexiondotorg:
@lvillani Thanks for contributing to the project :bow: I've been reflecting on the growing number of arguments in quickemu. I'm planning to remove most of the advanced configuration from arguments and expose it exclusively through options in the VM configuration. This is to simplify the code a little and also focus the command execution on common use cases.

So, I won't be merging this pull request as it stands, but if you want to refactor it to be automatically enabled when appropriate or exposed only via a configuration option, I'd welcome that :+1:

lvillani:
Thanks for the feedback! I completely agree with the goal of simplifying the arguments and expose this setting only in the VM configuration file instead. I'll update the PR to implement it that way in a few days.

lvillani:
I've updated the implementation so that viewer_extra_args is now exclusively read from the configuration file. I also realized I missed documenting this option in `quickemu_conf.5.md` earlier, so I've added that now.

I'm converting this PR to a draft while I perform more thorough testing. I'll mark it as ready for review once that's complete.

lvillani:
I have rebased my branch against `master` and set the PR ready for review. I'm a bit unsure about the way I documented `viewer_extra_args` in `quickemu_conf.5.md` (I wrote an example of how I use the option to set-up automatic USB redirection even though the option is general-purpose).

lvillani:
Rebased against the `master` branch and fixed commit ordering. The build now appears to fail for reasons unrelated to this PR.

lvillani:
@flexiondotorg Just bumping this since it's now ready for review! The changes reflect your feedback to move this directly to the VM configuration. Let me know what you think.

PR #1919, opened 2026-06-03
Update Debian & Archlinux for "sensible" 2026 defaults

funk-on-code:
Both Debian 12 and 13 can boot with both TPM and Secureboot enabled, as can Archlinux 2026.01 and beyond__(last tested with 2026.06.01)__, update the configurations to enable these.

# Description

When booting with a host with secureboot or TPM, Ubuntu Server and Debian and Archlinux will occasionally boot directly to PXE and break. Rather than manually editing configurations before first boot, I fixed __quickget__ to automatically apply these defaults instead.

There should probably be a current check for all OSes, I will be checking Ubuntu Desktop's and Tails in the next few weeks, as I have time.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] Packaging (updates the packaging)
- [ ] Documentation (updates the documentation)

# Checklist:

- [X] I have performed a self-review of my code
- [X] I have tested my code in common scenarios and confirmed there are no regressions
- [ ] I have added comments to my code, particularly in hard-to-understand sections __(none are required)__

### DIFF
--- a/quickget
+++ b/quickget
@@ -1673,6 +1673,7 @@ EOF
                     echo "tpm=\"on\"" >> "${CONF_FILE}"
                 fi
                 echo "secureboot=\"on\"" >> "${CONF_FILE}"
+                ;;
             deepin)
                 echo "disk_size=\"64G\"" >> "${CONF_FILE}"
                 echo "ram=\"4G\"" >> "${CONF_FILE}"
--- a/quickget
+++ b/quickget
@@ -1652,6 +1652,10 @@ EOF
         case ${OS} in
             alma|centos-stream|endless|garuda|gentoo|kali|nixos|oraclelinux|popos|rockylinux)
                 echo "disk_size=\"32G\"" >> "${CONF_FILE}";;
+            archlinux)
+                echo "secureboot=\"on\"" >> "${CONF_FILE}"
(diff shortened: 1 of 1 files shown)


alexhaydock:
If we start enabling Secure Boot by default I'd be keen to make it part of a larger review of compatible distros - if it's done for Arch & Debian then it should be done for Ubuntu, Alma, Rocky, CentOS, Fedora, etc.

Personally I'm in favour of this since I enable it for any VMs I run myself, but there's caveats which might run counter to Quickemu's goal of doing the right thing for most users:

- Not every host distro is going to ship the firmware with the Microsoft Platform Keys enabled to boot signed shims, so we'd need a fallback method for those platforms

- We would also need to call it out clearly in the documentation (in case anyone is actually trying to install DKMS packages etc inside the guests)

funk-on-code:
Fully agree @alexhaydock -- i've been going through them distro by distro as I have the time, and with things like fedora vs fedora core, there's a lot of updates to this I am interested in adding.

Once i'm done, i'll probably do the reverse too and put legacy boots and update RAM requirements too.

PR #1922, opened 2026-06-06
fix(quickget): emit curl progress and size to stdout when piped

### LINKED ISSUE #1921: When quickget is called from a GUI or script (stdout is a pipe, not a tty)
## Hello dear developers of Quickemu,

**Describe the bug**
When `quickget` is called from a GUI or script (stdout is a pipe, not a tty), `aria2c` suppresses its live progress output entirely. The caller receives no data until the download finishes — no percentage, speed, or ETA is available.
  
**To Reproduce**
  1. Call quickget with stdout piped: `quickget ubuntu 22.04 | cat`
  2. Observe that no aria2c progress lines appear during the download

**Expected behaviour** 
aria2c progress lines should be written to stdout even when piped, so that GUI frontends and scripts can parse real download progress. 
Passing `--summary-interval=1` to the aria2c invocation inside quickget would fix this — it forces a progress line every second regardless of tty:
  
  `[#3c3b01b03ef18480 512MiB/3.8GiB(13%) CN:1 DL:2.4MiB ETA:21m]`

This flag has no effect on normal terminal use where progress is already shown.
  
**Quickemu output**

  (no output until download completes — progress lines are suppressed when piped)

  
Running into this while building quickbox (https://github.com/musqz/quickbox), a GTK4/QT GUI for quickemu. 
Would really appreciate the fix — any GUI wrapper built on quickget hits this same wall. 

Thanks for developing QuickEmu!

musqz:
Description

  When quickget is called from a GUI or script (stdout is a pipe, not a TTY), curl's progress output is suppressed and the caller receives no feedback until
  the download finishes.

  This fix adds two behaviours when stdout is not a TTY:

  - A separate curl --head request fetches the correct total file size and emits it as content-length: N to stdout before the download starts. This avoids the
  ambiguity of --dump-header with redirects (multiple header blocks) or 206 resume responses (Content-Length = remaining bytes, not total).
  - --stderr - is added to the download curl call so the --progress-bar output is written to stdout, letting callers parse real-time percentage.

  Normal terminal behaviour is completely unchanged — PIPE_OPTS stays empty when stdout is a TTY.

  - Fixes #1921

  Type of change

  Checklist

### DIFF
--- a/quickget
+++ b/quickget
@@ -1464,7 +1464,17 @@ function web_get() {
         echo "- URL: ${URL}"
     fi
 
-    if ! curl --disable --progress-bar --location --output "${DIR}/${FILE}" --continue-at - --user-agent "${USER_AGENT}" "${HEADERS[@]}" -- "${URL}"; then
+    local PIPE_OPTS=()
+    if [ ! -t 1 ]; then
+        local CONTENT_LENGTH
+        CONTENT_LENGTH=$(curl --disable --silent --head --location \
+            --write-out '%{content_length_download}' --output /dev/null \
+            --user-agent "${USER_AGENT}" "${HEADERS[@]}" -- "${URL}" 2>/dev/null)
+        [ -n "${CONTENT_LENGTH}" ] && echo "content-length: ${CONTENT_LENGTH}"
+        PIPE_OPTS=("--stderr" "-")
+    fi
+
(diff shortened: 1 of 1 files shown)


musqz:
**Friendly ping** — just checking if this is on your radar. The CI failures (artixlinux, fedora, etc.) are upstream mirror flakiness ("N failed downloads"), unrelated to this change. 
Happy to rebase or make any adjustments if needed.

